PR #587, opened 2018-08-27
configure flask_caching to get rid of warning

ThomasWaldmann:
without configuring, it is disabled and whines about it (see the tests).

let's see if the 'simple' cache is good enough for development,
otherwise we'ld need to go to disk or silence the warning.


ThomasWaldmann:
have made a ticket pointing to here.

closing this as we can't merge it "as is" without causing these strange test failures.

PR #594, opened 2018-08-28
travis / tox: also test on pypy

ThomasWaldmann:
```
tox -e py27   # use this to only test on python 2.7
```
there are currently some test fails on pypy, have to look into that
later. on travis, allow failures for pypy.


ThomasWaldmann:
https://travis-ci.com/moinwiki/moin/jobs/142273259 py27 ok

https://travis-ci.com/moinwiki/moin/jobs/142273260 pypy with failures

ThomasWaldmann:
some of the fails are likely our fault (assuming some specific order for dicts/sets).

but the mimetype, sqlite and paramparser fail could be maybe pypy bugs / incompatilibities?

PR #604, opened 2018-08-29
tox: add flake8 for checking

ThomasWaldmann:
the ignores make the current sourcecode pass the checks.


ThomasWaldmann:
btw, i changed this from `pytest --pep8` to `flake8` to have the style checks separate from the functional tests.

PR #650, opened 2018-09-04
Update Brazilian Portuguese translation

rffontenelle:


ThomasWaldmann:
Thanks for updating this!

I'll do a first merge "as is", but guess I'll better update the english strings (see #651) and it would be nice if you could have a look after that again.

rffontenelle:
For the record, I've got confused when I found translation instructions in http://moinmo.in/MoinDev/Translation via web search, which are very different from https://moin-20.readthedocs.io/en/latest/devel/translate.html. I assume the latter is the current/correct one.  Maybe this could be avoided by adding an information box in the former to tell about newer instructions for newer versions.

ThomasWaldmann:
moin2 is very different in a lot of aspects, yes.

moinmo.in is a wiki, feel free to add a hint (if you do not have an account, write me an email with the username you would like to have).

PR #665, opened 2018-09-09
Converter tests

RogerHaase:
Fixed moinwiki_in_out tests by working around odd output when running under pytest.

Verified same odd output is generated under old hg repo.

Swapped the huge test based on moin 1.9 sample data to moin 2 sample data.

Did not remember the issue said to split the huge test into smaller pieces until I reread it after the commit, do you still want that done?

### DIFF
--- a/src/moin/converter/_tests/test_moinwiki_in_out.py
+++ b/src/moin/converter/_tests/test_moinwiki_in_out.py
@@ -19,8 +19,7 @@ from emeraldtree import ElementTree as ET
 from . import serialize
 
 from moin.util.tree import moin_page, xlink, xinclude, html
-# from moin.converter.moinwiki_in import Converter as conv_in
-from moin.converter.moinwiki19_in import ConverterFormat19 as conv_in
+from moin.converter.moinwiki_in import Converter as conv_in
 from moin.converter.moinwiki_out import Converter as conv_out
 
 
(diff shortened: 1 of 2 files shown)


src/moin/converter/_tests/test_moinwiki_in_out.py
```diff
@@ -19,8 +19,7 @@
 from . import serialize
 
 from moin.util.tree import moin_page, xlink, xinclude, html
-# from moin.converter.moinwiki_in import Converter as conv_in
-from moin.converter.moinwiki19_in import ConverterFormat19 as conv_in
+from moin.converter.moinwiki_in import Converter as conv_in
 from moin.converter.moinwiki_out import Converter as conv_out
```
ThomasWaldmann: hmm, was the intention of this test to test the 1.9 -> 2.0 conversion?

is `moinwiki19_in` and `moinwiki_out` the stuff we use for upgrade?

ThomasWaldmann:
src/moin/converter/moinwiki_out.py:345:9: E722 do not use bare except'

src/moin/converter/_tests/test_moinwiki_in_out.py
RogerHaase: There is a test_moinwiki19_in_20_out.py to test the 1.9 -> 2.0 conversion.

m import uses moinwiki19_in and moinwiki_out to convert the most recent wiki page to 2.0 format. The older pages are left in 1.9 format.

src/moin/converter/moinwiki_out.py
```diff
@@ -340,10 +340,14 @@ def open_moinpage_note(self, elem):
     def open_moinpage_nowiki(self, elem):
         """{{{#!wiki ... or {{{#!highlight ... etc."""
         nowiki_marker_len, all_nowiki_args, content = elem._children
-        nowiki_args = all_nowiki_args[0]
+        try:
+            nowiki_args = all_nowiki_args[0]
+        except:
```
ThomasWaldmann: `except IndexError:` to avoid flake8 complaining.

PR #667, opened 2018-09-13
update gitignore; remove moin (matched /moin/), fixed messages.mo, ad…

RogerHaase:
…ded upload.py and comments

removed moin as it matched all files below /moin/ directory, hiding error in messages.mo. 

^moin$ was in hgignore; seems like an error - if there were such a file, wouldn't it override moin in virtualenv?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ docs/devel/api/
 instance/
 MANIFEST
 src/moin/_version.py
-src/moin/translations/.*/LC_MESSAGES/messages.mo
+src/moin/translations/*/LC_MESSAGES/messages.mo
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -23,13 +23,15 @@ wikiconfig_*.py
...
 activate.bat
 deactivate.bat
 moin.bat
 m.bat
 activate
-moin
 m
+# files created by make.py
 m-*.txt
+upload.py
```
ThomasWaldmann: `upload.py` is the rietveld (google codereview) uploader. I removed related code and guess we we won't use that again now we can do code reviews on github PRs.

RogerHaase:
removed upload.py

PR #688, opened 2018-09-22
Quickinstall

RogerHaase:


quickinstall.py
```diff
@@ -104,9 +104,9 @@ def do_install(self):
         if pip_version < [6, 0]:
             args += ['--download-cache', self.download_cache, ]
         if pip_version >= [1, 5]:
-            args += [
-                '--process-dependency-links',
-            ]
+            args += ['--process-dependency-links', ]
```
ThomasWaldmann: do we need this at all? iirc i removed all the hg/git repo links from setup.py.

quickinstall.py
```diff
@@ -104,9 +104,9 @@ def do_install(self):
...
+        if pip_version >= [9, 0]:
+            args += ['--upgrade-strategy=eager', ]
```
ThomasWaldmann: https://github.com/pypa/pipenv/issues/438 9.0 or 10.0?

ThomasWaldmann:
also it upgraded `Markdown` from 2.6.x to 3.0 and this broke a lot of tests.

quickinstall.py
RogerHaase: https://github.com/girder/girder/pull/2695 

manthey commented on Apr 17

pip 9.x honors the flag. pip 8 fails if it is specified.

jbeezley commented on Apr 17

The flag was added in 9, the default was changed in 10.
--

PR #690, opened 2018-09-23
Markdown 3.0.0 fails our converter tests, require <3.0.0 for now

ThomasWaldmann:


ThomasWaldmann:
@RogerHaase update your master and rebase your PR onto that, should fix it.

PR #706, opened 2018-10-07
fix #691 add changes for markdown 3.0.0

### LINKED ISSUE #691: Markdown 3.0.0 fails converter tests
#690 just requires <3.0.0 for now, but we should check why that is and whether we should fix it or report an issue to Markdown project.

RogerHaase:


src/moin/converter/markdown_in.py
```diff
@@ -58,10 +58,12 @@ def postproc_text(markdown, text):
     if text is None:
         return None
 
-    for pp in markdown.postprocessors.values():
+    for pp in markdown.postprocessors:
         text = pp.run(text)
 
     if text.startswith('<pre>') or text.startswith('<div class="codehilite"><pre>'):
+        # with markdown 3.0.0 indented code is somehow escaped twice
+        text = text.replace('&amp;lt;', '&lt;').replace('&amp;gt;', '&gt;').replace('&amp;amp;', '&amp;')
```
ThomasWaldmann: `werkzeug.unescape(text)` maybe?

src/moin/converter/_tests/test_markdown_in.py
```diff
@@ -39,7 +39,7 @@ def setup_class(self):
         (u'<http://moinmo.in/>',
          '<p><a xlink:href="http://moinmo.in/">http://moinmo.in/</a></p>'),
         (u'[yo](javascript:alert("xss"))',
-         '<p>javascript:alert("xss")</p>'),
+         '<p>javascript:alert()</p>'),
```
ThomasWaldmann: where did the "xss" go?

RogerHaase: it now comes across as a title attribute and is ignored. That is a rather strange test, old output was malformed, new output is also malformed

src/moin/converter/markdown_in.py
RogerHaase: I was using that at first, but then worried it may do too much. I wanted to reverse just these 3 strings from markdown serializers.py:

    def _escape_cdata(text):
        # escape character data
        try:
            # it's worth avoiding do-nothing calls for strings that are
            # shorter than 500 character, or so.  assume that's, by far,
            # the most common case in most applications.
            if "&" in text:
                # Only replace & when not part of an entity
                text = RE_AMP.sub('&amp;', text)
            if "<" in text:
                text = text.replace("<", "&lt;")
            if ">" in text:
                text = text.replace(">", "&gt;")
            return text
        except (TypeError, AttributeError):  # pragma: no cover
            _raise_serialization_error(text)

If you prefer the workzeug, I will change

ThomasWaldmann: huh, strange...

PR #711, opened 2018-10-12
fix #679 creole_in: remove trailing whitespace in cells

### LINKED ISSUE #679: creole_in: remove trailing whitespace in cells?
while fixing #386, i noticed that leading ws in table cells is stripped, but trailing ws is not:

```
| text |
```
leads to something like:
```
<table-cell>text </table-cell>
```

RogerHaase:


src/moin/converter/creole_in.py
```diff
@@ -562,7 +562,7 @@ def tablerow_cell_repl(self, stack, cell, cell_text, cell_head=None):
             attrib[moin_page.class_] = 'moin-thead'
         element = moin_page.table_cell(attrib=attrib)
         stack.push(element)
-        self.parse_inline(cell_text, stack)
+        self.parse_inline(cell_text.strip(' '), stack)
```
ThomasWaldmann: hmm, just `.strip()` would have stripped all whitespace. was it intended to only strip blanks?

PR #727, opened 2018-10-13
move dev scripts, fixes #638

### LINKED ISSUE #638: move developer scripts
Looks like these are not used within moin, just by developers:

e.g.:
- src/moin/script/win/*.py
- contrib/pep8/coding_std.py
- contrib/wsgi/raw_wsgi_bench.py
- quickinstall.py
- make.py (?)

move to:
- scripts/...

ThomasWaldmann:
didn't touch quickinstall / make yet.


ThomasWaldmann:
@RogerHaase is there some need that `make.py` and `quickinstall.py` are 2 scripts?

Because quickinstall relies on being able to `import make`, both have to be in `.` right now.

If both were put into 1 file: no need to import and I could just move them to `scripts/`.

Entering `python scripts/make.py quickinstall` should also be no problem for rare usage.

ThomasWaldmann:
@RogerHaase to answer your irc comments: this pr is not related to pypi, rather about putting some scripts that are only intended for developers (working in a git checkout) to a specific place, removing clutter from the toplevel dir.

Such scripts that are unrelated to running moin in production (e.g. wget.py and dos2unix.py) do not need to be in the moin python package (as `moin.x.y`), neither in the pypi package.

What I found confusing about make.py and quickinstall.py is that both use/call each other by different means: by import or by creating a subprocess. As usually only `.` will be in `sys.path`, but not `./scripts`, the import blocks moving both to `scripts/` dir.

So I guess best is to defer dealing with make/quickinstall to separate PRs.

ThomasWaldmann:
I think we should only have stuff for developer's convenience in make/quickinstall scripts.

Everything a normal user (moin admin) needs after installation needs to be moved into `moin.scripts` (and be available as `moin ...` commands). This does not need to include installation because this often will be handled elsewhere (e.g. linux package manager) or the admin will know it.

PR #733, opened 2018-10-25
fixes #731, ajaxdestroy fails to destroy/delete files with non-ascii names

### LINKED ISSUE #731: ajaxdestroy fails to destroy/delete files with non-ascii names
Given a new wiki loaded with sample data, go to Global Index, click Select All button, then Destroy button. 

There will be a message saying 23 items destroyed, 1 not destroyed. There resulting Global Index view shows all files names except TemplateSample were removed from the prior display. The TemplateSample should have been destroyed.

Reloading the Global Index view shows all files with non-ascii names and the TemplateSample remain.

If the subitems of OtherTextItems is viewed, Select All button is clicked, then Destroy button is clicked there is a similar failure. A message saying 5 items were destroyed is shown, but none were destroyed.

 There are similar results if the Delete button were clicked. 

This failure happens on both Windows and *nix.

RogerHaase:
ajaxdelete and ajaxdestroy have redundant code

### DIFF
--- a/src/moin/apps/frontend/views.py
+++ b/src/moin/apps/frontend/views.py
@@ -22,6 +22,7 @@ import uuid
 import mimetypes
 import json
 import threading
+import urllib
 from io import BytesIO
 from datetime import datetime
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


src/moin/apps/frontend/views.py
```diff
@@ -898,8 +899,15 @@ def ajaxdelete(item_name):
         for itemname in itemnames:
             response["itemnames"].append(itemname)
             itemname = subitem_prefix + itemname
+            if '%' in itemname:
```
ThomasWaldmann: if it is correct with that "if" it should work in the same way even if the condition is not true.

src/moin/apps/frontend/views.py
```diff
@@ -924,8 +932,15 @@ def ajaxdestroy(item_name):
...
```
ThomasWaldmann: same here.

PR #741, opened 2018-10-29
Patches

RogerHaase:


src/moin/static/js/index_action.js
```diff
@@ -14,11 +14,10 @@ $("document").ready(function () {
     var POPUP_FADE_TIME = 200, // fade in, fade out times for selected popups
         IFRAME_CREATE_DELAY = 200, // delay between start of multiple downloads
         IFRAME_REMOVE_DELAY = 3000, // life expectancy of iframe used for file downloads
-        MESSAGE_VIEW_TIME = 4000, // life expectancy of delete/destroy status messages
         // delete and destroy process started and completed messages
         ACTION_LOADING = {'delete': _("Deleting.."), 'destroy': _("Destroying..")},
         ACTION_DONE = {'delete': _("Items deleted: "), 'destroy': _("Items destroyed: ")},
-        ACTION_FAILED = {'delete': _(", Items not deleted: "), 'destroy': _(", Items not destroyed: ")};
+        ACTION_FAILED = {'delete': _(", Not authorized, items not deleted: "), 'destroy': _(", Not authorized, items not destroyed: ")};
```
ThomasWaldmann: is a auth fail the only possible reason for the action failing?

RogerHaase: There may be more bugs causing failures, but that possibility should not be included in the message. If there is a network or hardware failure, I hope those are handled upstream so the user sees Internal Server Error or similar.

The Global Index shows all items that allow view, so lack of authorization for delete/destroy is possible.

I can not think of another cause for failure.

ThomasWaldmann: ok, thanks.

PR #754, opened 2018-11-05
Patches

RogerHaase:


src/moin/translations/ru/LC_MESSAGES/messages.po
```diff
@@ -181,7 +181,7 @@ msgstr ""
 #: src/moin/apps/admin/views.py:83
 #, fuzzy, python-format
 msgid "User profile of %(username)s: %(email)s %(disabled)s"
-msgstr "Профиль пользователя %(username)s: %(email)r"
+msgstr "Профиль пользователя %(username)s: %(email)s"
```
ThomasWaldmann: disabled placeholder is still missing.

ThomasWaldmann: good that you let that "fuzzy" there. likely some strings are still wrong, it just makes the warnings at install time go away.

PR #765, opened 2018-11-15
fix #579 correct parentid for revert, rename, delete, convert Xactions

### LINKED ISSUE #579: meta.parentid may contain incorrect data
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/579) by RogerHaase (Bitbucket: [RogerHaase](https://bitbucket.org/RogerHaase), GitHub: [RogerHaase](https://github.com/RogerHaase)).**

----------------------------------------

If one creates a new item, the initial meta.parentid is None. When the item is modified, each successive revision has a meta.parentid that points to the parent, or prior revision. All is well.

However, if one views the item history of an item with several revisions and reverts one of the old versions, the meta.parentid of the new current revision is not updated. If the first revision is reverted, the meta.parentid is None. If a middle revision is reverted, the current meta.parentid continues to point to the prior parent.

If one views the item history of an item and destroys a revision other than the current revision, then the child revision of the destroyed revision has a meta.parentid that is invalid as it points to a revision that no longer exists.

If an administrator runs the command "moin maint-reduce-revisions", the meta.parentid is not set to None, so all items that had multiple revisions will have an invalid meta.parentid.

This needs rechecking, but a guess is that "moin item-get" and "moin item-put" have similar issues.

Many of the items loaded from the "m sample"  examples have meta.parentid values that point to non-existent revisions.

Because of these issues, attempts to run a diff from the Global History view will result fail is various ways.

    . 404
    . traceback
    . create new item

A workaround for the diff crashes was added: [f03e8e1 (bb)](https://bitbucket.org/thomaswaldmann/moin-2.0/commits/f03e8e1)

RogerHaase:


src/moin/items/__init__.py
```diff
@@ -674,7 +674,8 @@ def _save(self, meta, data=None, name=None, action=ACTION_SAVE, contenttype_gues
             meta[REV_NUMBER] = 1
 
         if not overwrite and REVID in meta:
-            # we usually want to create a new revision, thus we must remove the existing REVID
+            # we usually want to create a new revision, thus we update parentid and remove the existing REVID
+            meta[PARENTID] = currentrev.meta[REVID] if currentrev else meta[REVID]
             del meta[REVID]
```
ThomasWaldmann: looks good for the case of an existing ("on disk") currentrev.

but what happens if currentrev is falsy - then we keep its revid as parentid, kill the revid from meta (which causes the storage to generate a new revid, right?). what is the meaning of the parentid then?

RogerHaase: The only case I could find where currentrev is false is for a new item and then "REVID in meta" is also false so meta[PARENTID] is not set because the if fails.

I think currentrev will always be true when the if succeeds and I could not find a case where it was not true. The fallback to meta[REVID] was to avoid a traceback just in case.

ThomasWaldmann: we can keep that as is for now.

PR #774, opened 2018-11-24
Patches

RogerHaase:


src/moin/apps/frontend/views.py
```diff
@@ -1117,6 +1117,8 @@ def name_initial(files, uppercase=False, lowercase=False):
             title = _("Index of subitems '%(item_name)s'", item_name=item_name)
     else:
         title = _("Global Index")
+    if isinstance(item.rev.data, file):
+        item.rev.data.close()
```
ThomasWaldmann: hmm, maybe it would be more flexible to check for a `close` attribute and then calling it. instead of a `file` it could be also something file-like, like a `BytesIO`.

RogerHaase: Windows only objects to real files left open in another process, so bytesio and stringio object may be left open. I will convert the many existing tests to a function so it may be easily changed.

ThomasWaldmann: sure, they are not directly causing issues.

but if we can / want to close a file, we don't need to let open file-like objects either.

Issue #98, opened 2011-11-05
fix PARENTID usage, use MTIME less


ThomasWaldmann:
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/98) by Thomas Waldmann (Bitbucket: [thomaswaldmann](https://bitbucket.org/thomaswaldmann), GitHub: [thomaswaldmann](https://github.com/thomaswaldmann)).**

----------------------------------------

some views use MTIME based sorting of revisions, but they should rather use the order of revisions that is defined by the revisions' PARENTID pointer.

also import19 (1.9 -> 2.0 importer) doesn't create PARENTID metadata at all.

RogerHaase:
afaik the PARENTID chain is now correct except when an item's revision is destroyed thereby breaking the chain.

The original report suggests MTIME sorting may be incorrect or inferior to a PARENTID sorting. Is there an example when this may be true. Sorting by the PARENTID chain seems cumbersome.

ThomasWaldmann:
for normal, linear history, mtime sorting should be fine (assuming monotonic time).

if there would be forks in history (which we do not generate, iirc), it would make a difference.

ThomasWaldmann:
if one just has a list of revision, mtime sort is easier than by-parent-pointers.

otoh, if you have a specific revision and want to know it's parent, obviously using the parentid "pointer" is easier than trying to find out via timestamps.

RogerHaase:
As stated above, PARENTID is correct, except in the case when an item revision is destroyed. There is no failing transaction needing this functionality. 

In addition to MTIME, there is REV_NUMBER in metadata. 

Have not tried, but I think a search for rev_number > n, sorted, limit 1, would be an acceptable search alternative should there be a need to find a parent or closest ancestor revision. 

A transaction needing a parent revision my want to know the parent was destroyed; a broken PARENTID link after a revision is destroyed may be the correct state.

`Note parentid is ambiguous, in some places (hierarchic ACL processing) there is code using parentid to refer to a subitem's parent.`

Lowered priority and removed milestone. Keeping this open as there is some code in /apps/feed/views.py that uses PARENTID.

UlrichB22:
Can this issue be closed?
There was a duplicate issue #1448.

RogerHaase:
yes, thanks for finding it.

PR #780, opened 2018-12-07
add support for PARENTID in import19.py, partial fix for #98

RogerHaase:


src/moin/scripts/migration/moin19/import19.py
```diff
@@ -235,10 +236,15 @@ def iter_revisions(self):
             fnames = os.listdir(revisionspath)
         except OSError:
             fnames = []
+        parent_id = u''
         for fname in fnames:
             try:
                 revno = int(fname)
-                yield PageRevision(self, revno, os.path.join(revisionspath, fname))
+                page_rev = PageRevision(self, revno, os.path.join(revisionspath, fname))
+                page_rev.meta[PARENTID] = parent_id
```
ThomasWaldmann: for the first revision, this will have parent_id == u'' - is this how it is handled at other places also or do we have a None value or no metadata entry at all?

PR #813, opened 2019-04-09
ItemPageList updates per request

kwatsen:
After merging the PR for issue #798, Roger suggested a number of improvements (e.g., removing the numsep parameter, translations, test for non-existence, etc.), all of which are addressed here.  Please take a look at the new sample/MoinWikiMacros page!

I'm waiting a couple more weeks before removing the <<PageNameList>> macro.

One remaining enhancement is to add filtering by tags as yet another filtering mechanism.

### DIFF
--- a/contrib/sample/MoinWikiMacros.data
+++ b/contrib/sample/MoinWikiMacros.data
@@ -321,7 +321,7 @@ An unordered list links for all descendents of the "Foo/Bar" item, using
 the un-cameled text of each decendent's name, using a '-' character between
 blocks of letters and blocks of numbers.
 
-<<ItemPageList("Foo/Bar", display="UnCameled", numsep='-')>>
+<<ItemPageList(item="Foo/Bar", display="UnCameled", numsep='-')>>
 }}}
 
 '''Result:'''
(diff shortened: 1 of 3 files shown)


RogerHaase:
Nitpicking on MoinWikiMacros.data:

You went from heading level 3 to level 5 several times.

 === ItemPageList ===
+===== The "item" parameter =====

On ItemPageList:

Translations like this, where you want the variable inserted in the string and
the value of the variable is not available until runtime:

raise KeyError(_('Unrecognized key "%s".' % key))

are better formatted like this curious lazy form (leading _ in _namespace is not required):

 = L_("%(_namespace)s is not a valid namespace.", _namespace=current_namespace)

so translator sees something like this in message.po file:

#: src/moin/forms.py:169
#, python-format
msgid "%(_namespace)s is not a valid namespace."
msgstr "%(_namespace)s ist kein gÃ¼ltiger Namespace."

In above, translator sees variable name and can write phrase based on the hint. He can change position of variable name within text string or use tense, male/female phrasing, etc.

I am English-only speaker. The above is my best attempt to explain it.

PR #818, opened 2019-04-13
fix moin2.wsgi script

kwatsen:
The current script was 3 years old ands referenced the "MoinMoin" app that doesn't exist anymore.

Also cleaned up some of the comments.


RogerHaase:
If you start the commit message with "fixes #818" then this issue would have been closed automatically and future devs looking at the change log could learn more about the need for this patch.

PR #821, opened 2019-04-15
Dokunized theme

kwatsen:
Don't merge this PR!   I'm using the PR to more easily show off what I've done so far (a cumulative diff)...

Actually, this isn't the theme I wanted to create initially and, in fact, I renamed the theme along the way to better reflect what it is.   Whether it ever gets merged into master to open to discussion (especially since I mimics another's wiki's default look-n-feel.

FWIW, the theme I wanted to write was going to have a default "dark" mode, with a user-selectable light mode.  I was unable to do this because so many color settings are being made in the default templates and CSS files.

Also, I should add, that the `theme_supp` (class ThemeSupport) was/is *much* more helpful than `utils.html`.  Manipulating Python objects in Jinga is much easier that manipulating HTML text.  

Suggest making more theme_supp functions, hopefully eliminating the need to used utils.html.  As for colors, also suggest making a color-pallet (e.g., 10-15 colors) that the default is mapped to, but enables themes to easily target the colormap used.


ThomasWaldmann:
I didn't look at the code yet nor at how the theme looks like, but wanted to tell about a general problem with theme development:

1. Often users find the moin themes somehow ugly, missing some functionality or otherwise not what they wanted. I agree with that.
2. Then they start their own all new theme, that is somehow better in some respect (but maybe also worse in others).
3. Now moin has N+1 sub-optimal themes.
4. Goto 1

I'ld appreciate if we do NOT grow the number of themes, but work together and concentrate work on improving a few.

I don't care if these few ones are the ones we already have or overall better new ones though.

kwatsen:
I agree.  

This theme is what I came up with without changing any internal files (e.g., base.html, theme_supp, updated font awesome, bootstrap, etc.).  I will improve it more over time (e.g., fix mobile view, CSM-like presentation for read-only users, etc.).  This theme solves my immediate goal of having a better theme.   

Before, when using 1.9, I tried 10+ themes, and found only a couple that were passingly okay.  I don't understand the history of the current default themes, nor if there are any requirements to maintain a particular theme, versus replacing them all with just *one* new default theme.

Having one new default theme that doesn't thouch any internal files is the easy route.   Of course, better results will come if we dig deeper.  For instance, we could:

  * Update font-awesome to the latest version and use the highlighted version of the icons when hovering.  
  * Or, better, enable themes to define all the CSS used (not just extra CSS), then they themselves could update the version of font-awesome used.
  * Define a site-specific local config knob to specify if CDNs or Moin should serve static (CSS, JS, etc.) files when CDN-alternates exist.
  * Enable internal Moin-generated HTML (e.g., per templates, etc) to have a color palate that can be [re-]defined by the theme.   This is just an idea, what whatever works to support a "dark" mode.

This sounds like the right thing to do, and a lot of work.  Perhaps time is better spent elsewhere, to get 2.0 out the door?

RogerHaase:
This cannot be merged due to many changes in other moin code. Better to work on improving other existing themes.

kwatsen:
Did you ever try it?

Can I do a screen-share with you so you can see it?

RogerHaase:
kwatsen lives! Hi!

I have the Dokunized theme running. iirc I fixed a few small bugs long ago but I don't remember what they were.   Attached is a zipfile of the current state on my PC. But per TW's comment above we should have a limited number of default themes.  

If you want to continue development, maybe add it to: 

    http://moinmo.in/ThemeMarket

As a next step and see what happens... 

[dokunized.zip](https://github.com/moinwiki/moin/files/10421055/dokunized.zip)

kwatsen:
Excellent, I'm glad to hear that you tried it out.   I guess it wasn't compelling enough to make the default theme?   ...because, you know, the existing default is really dated  ;)

Yes, I've been away for awhile, as I tend to only work on such projects over Winter, when I rather stay inside (the ground not being warm enough to garden).   I am, in fact, about to pickup on that same project again in a few weeks...

PR #838, opened 2019-06-21
Sql

RogerHaase:


src/moin/utils/show_time.py
```diff
@@ -0,0 +1,69 @@
...
+        return _("minutes"), (seconds+30)//60
+    if seconds < 129600:  # 36 hours
+        return _("hours"), (seconds+1800)//3600
+    if seconds < 864000:  # 10 days
+        return _("days"), (seconds+43200)//86400
+    if seconds < 4838400:  # 8 weeks
+        return _("weeks"), (seconds+302400)//604800
+    if seconds < 63072000:  # 24 months
+        return _("months"), (seconds+1296000)//2592000
+    return _("years"), (seconds+15768000)//31536000
```
ThomasWaldmann: I didn't check where this is called, but in general, I really hate such relative values.

Not quite infrequently, I want to find out at what specific point in time something happened and some web site tells me "42 days ago" or something, forcing me to manually do calendar computations just to know the datetime which it could have just told me right from the beginning.

And even worse: precision goes down with time, so with "2 years ago", nobody can compute when it happened.

src/moin/utils/show_time.py
```diff
@@ -0,0 +1,69 @@
...
+    """Return current or passed (dt) time in user's preferred format or default to moin-style ISO 8601."""
+    if utc_dt is None:
+        utc_dt = datetime.datetime.utcnow()
+    user_tz = i18n.get_timezone()
+    if user_tz:
+        tz = pytz.timezone(user_tz)
+        fmt = '%Y-%m-%d %H:%M'
+    else:
+        tz = pytz.utc
+        fmt = '%Y-%m-%d %H:%MZ'
```
ThomasWaldmann: I'ld define fmt just once and then append "Z" or not.
That way, one only needs to update 1 place when changing it.
Also, how about seconds?

src/moin/utils/show_time.py
```diff
@@ -0,0 +1,69 @@
...
+    loc_dt = tz.localize(utc_dt)
+    dt = loc_dt.strftime(fmt)
+    print '%s' % dt
```
ThomasWaldmann: debug code left over?

src/moin/utils/edit_locking.py
```diff
@@ -0,0 +1,331 @@
+# Copyright: 2019 MoinMoin:RogerHaase
+# License: GNU GPL v2 (or any later version), see LICENSE.txt for details.
+
+"""
+Create sqlite3 data models for edit locking and saving drafts. Expected usage
```
ThomasWaldmann: I have a django project which uses sqlite3 db in production and, sadly, it sucks.

I have to daily clean up my inbox from all the "Database is locked" exceptions which occur when there are too many parallel writers.

I tried to find a way to fix this, but it seems it is a design limitation and the only fix is to use a real db like postgresql.

contrib/loadtesting/locust/locustfile.py
```diff
@@ -158,7 +164,7 @@ def create_home_page(self):
         home_page_post = '/+modify/users/' + self.user_name + '?contenttype=text%2Fx.moin.wiki%3Bcharset%3Dutf-8&itemtype=default&template='
         response = self.client.post(home_page_post,
                                     {"content_form_data_text":new_content,
-                                     "comment":"yes",
+                                     "comment":"my comment",
```
ThomasWaldmann: pep8 nitpick: add a blank after the `:`

contrib/loadtesting/locust/locustfile2.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+user_number = 0
+
+
+def get_textarea(html):
+    """Return contents of textarea where html is html output from +modify"""
+    html = html.split('<textarea ')[1]
+    html = html.split('>')[1]
+    html = html.split('</textarea>')[0][:-10]
```
ThomasWaldmann: hehe, html parsing from hell. :-)

contrib/loadtesting/locust/locustfile2.py
```diff
@@ -0,0 +1,207 @@
...
+    return html
+
+
+def format_date_time(dt=None):
+    """Return current or passed (dt) time in a human readable format."""
+    if dt is None:
+        dt = datetime.datetime.now()
+    fmt = '%Y-%m-%d %H:%M:%S'
+    dt = dt.strftime(fmt)
+    print '%s' % dt
```
ThomasWaldmann: debug print?

contrib/loadtesting/locust/locustfile2.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @seq_task(3)
+    def click_post_register(self):
+        response = self.client.post("/+register",
+                                    {"register_username":self.user_name,
+                                     "register_password1":"locust123",
+                                     "register_password2":"locust123",
+                                     "register_email":self.user_email,
+                                    })
+        if not response.status_code == 200:
```
ThomasWaldmann: != (multiple places with this issue)

src/moin/utils/show_time.py
RogerHaase: In general I agree, but in this case the duration will be a few seconds or minutes in the vast majority of cases. In that case, giving a full date and time seemed too verbose. 

Maybe a compromise is to change the message to use a date-time if the interval is over 36 hours or so,

src/moin/utils/edit_locking.py
RogerHaase: Certainly a concern and I worry about bugs. Using the Locust load test with 5 locusts trying to update Home, there are a lot of successful failed attempts to edit.

Do you use:

        self.con.isolation_level = 'EXCLUSIVE'
        self.con.execute('BEGIN EXCLUSIVE')

on your app? That is supposed to lock the sqlite database until a commit or close. I think at the OS level there are 10 tries 1 second apart, then a failure. It is on my list to add a check for such a failure.

src/moin/utils/show_time.py
RogerHaase: Trying to save 3 characters on some wide reports. Seconds seem overly precise.

src/moin/utils/edit_locking.py
ThomasWaldmann: "there are a lot of successful failed attempts" - did you mean "a lot of failed attempts"?

in the django app, the ORM is talking to sqlite and I tried some tweaking, but it did not really get better.

src/moin/utils/edit_locking.py
```diff
@@ -0,0 +1,331 @@
...
+            u_name, i_id, i_name, rev_number, save_time, i_rev_id = draft
+            if i_id == self.item_id:
+                if not overwrite:
+                    return
+                # in case of timeout and update by someone else get rev_number from prior Preview
+                draft_rev_number = rev_number  # XXX per line 1074 in __init__
+                rev_id = i_rev_id
+                self.draft_name = self.make_draft_name(rev_id)
+            self.cursor.execute('''DELETE FROM editdraft WHERE user_name = ? ''', (self.user_name, ))
+        if data_in:
```
ThomasWaldmann: hmm, guess such kind of code is sqlite3 specific and if we run into troubles with it (like I did with the django app), changing to something else is not even simple (with the django ORM, I can also use pgsql - that's at least the theory).

ThomasWaldmann: I also talked to some fellow python guys at last europython (or pycon.de) and they recommended always using a ORM and not directly talking to the DB implementiation (I asked about pg, but guess the same applies to sqlite3, maybe even more due to the scalability / parallel writers issues).

kwatsen: SQLAlchemy supports an API between ORM and raw-queries.  For example:

```
metadata = sa.Metadata((bind=engine)
metadata.reflect()
t = sa.Table(table_name, self.metadata, autoload=True)
with engine.connect() as conn
    result = conn.execute(
        sa.select([t.c.id]).where(r.c.name == _name)
    )
```

I think this syntax is interpreted according for the underlying SQL engine used.

RogerHaase: See issue https://github.com/moinwiki/moin/issues/839

It has been ~7 years since I last wrote any sqlalchemy related code. I would welcome any help you could give.  I have no plans to work on https://github.com/moinwiki/moin/issues/839 soon. 

There are Locust load tests in /contrib/loadtesting/locust.

PR #847, opened 2019-07-13
Add shebang line to `./m` on non-windows

### LINKED ISSUE #846: ./m command does not run when using fish shell
When using the [fish shell](https://fishshell.com/) on Debian, the `./m` command fails as it does not guess that the file should be executed with `sh` or `bash`.

I suggest adding a shebang referring to `/bin/sh` as a solution.

hoh:
Add shebang line to `./m` on non-windows. Fixes #846 

This prevents it from failing when using non-bash shells.


ThomasWaldmann:
thanks!

PR #867, opened 2019-08-04
fix #864 Inconsistent line endings in /wiki/data/default/data/: \r\n …

### LINKED ISSUE #864: Inconsistent line endings in /wiki/data/default/data/: \r\n vs. \n
Per https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4: "As with all MIME transmissions, "CR LF" (i.e., `%0D%0A') is used to separate lines of data." e.g. lines of data pulled from a form's textarea are separated by \r\n. 

If a text item is edited and saved, the saved file in /wiki/data/default/data/ contains \r\n as line separators.

I believe the above is the desired state.

However, the current "moin get-item ..." incorrectly replaces \r\n with \n in the .data output. As get-item is used to extract the items in /contrib/sample/, these files have incorrect \n line endings. These incorrect line endings then persist when the sample data is loaded on a wiki. 

After the py3 conversion, one adverse result is that the code near line 40 in /scripts/main/modify_item.py:

    data = rev.data.read().replace('\r\n', '\n')

fails. A solution is to remove the `.replace('\r\n', '\n')`.  The moin item-put  command accepts the \r\n line endings as input and retains \r\n when storing the item.

As for the meta data, on windows the meta data is stored with \r\n line endings. *nix systems likely will save data with \n line endings (not yet tested). Both are acceptable as moin item-put stores meta data in Jason format; all line endings are removed.

TODO:
 * fix #862, correcting line endings on item-get, item-put
 * recreate all /contrib/sample items with correct \r\n line endings
 * retest /utils/diff3.py creates incorrect output after edit conflict

RogerHaase:
…vs. \n

reloading sample data will create \r\n line endings for text data in
/wiki/data/default/data

### DIFF
--- a/src/moin/scripts/maint/modify_item.py
+++ b/src/moin/scripts/maint/modify_item.py
@@ -6,6 +6,7 @@ MoinMoin - get an item revision from the wiki, put it back into the wiki.
 """
 
 import json
+import io
 
 from flask import current_app as app
 from flask_script import Command, Option
@@ -35,6 +36,17 @@ class GetItem(Command):
         meta = json.dumps(dict(rev.meta), sort_keys=True, indent=2, ensure_ascii=False)
         with open(meta_file, 'w', encoding='utf-8') as mf:
             mf.write(meta)
+
+        if 'charset' in rev.meta['contenttype']:
+            # input data will have \r\n line endings, output will have platform dependent line endings
+            charset = rev.meta['contenttype'].split('charset=')[1]
+            data = rev.data.read().decode(charset)
+            lines = data.splitlines()
+            lines = '\n'.join(lines)
+            with open(data_file, 'w', encoding=charset) as df:
+                df.write(lines)
+            return
+
         data = rev.data.read()
         with open(data_file, 'wb') as df:
             df.write(data)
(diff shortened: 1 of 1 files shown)


src/moin/scripts/maint/modify_item.py
```diff
@@ -35,6 +36,17 @@ def run(self, name, meta_file, data_file, revid):
...
+
+        if 'charset' in rev.meta['contenttype']:
+            # input data will have \r\n line endings, output will have platform dependent line endings
+            charset = rev.meta['contenttype'].split('charset=')[1]
+            data = rev.data.read().decode(charset)
+            lines = data.splitlines()
+            lines = '\n'.join(lines)
+            with open(data_file, 'w', encoding=charset) as df:
+                df.write(lines)
+            return
```
ThomasWaldmann: maybe open a ticket to refactor this in a general-use utility function. it will likely be needed at a lot of places.

(ticket because some of the recent changes in the py3 branch are NOT related to py3 porting, but rather are generic development activity)

src/moin/scripts/maint/modify_item.py
```diff
@@ -67,5 +79,21 @@ def run(self, meta_file, data_file, overwrite):
...
         item = app.storage.get_item(**query)
+
+        # we want \r\n line endings in data out because \r\n is required in form textareas
+        if 'charset' in meta['contenttype']:
+            charset = meta['contenttype'].split('charset=')[1]
+            with open(data_file, 'rb') as df:
+                data = df.read().decode(charset)
+            if '\r\n' not in data and '\n' in data:
+                data = data.replace('\n', '\r\n')
+                data = data.encode(charset)
```
ThomasWaldmann: same for this.

PR #893, opened 2019-08-08
Update to python3.6

eamanu:
Bump to py3.6. This is in WIP but the first step to bump to py3.6

I update the quickinstall.py. Remove the virtualenv requirements to start use
venv.

### DIFF
--- a/quickinstall.py
+++ b/quickinstall.py
@@ -9,7 +9,7 @@ Usage for installation:
 
     <python> quickinstall.py (where <python> is any Python 3.5+ executable)
 
-Requires: Python 3.5+, virtualenv, pip
+Requires: Python 3.6+, pip
 
 After initial installation, a menu of frequently used commands is provided for
 moin2 developers and desktop wiki users.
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
See the issue about the py36 upgrade. Qubes OS is still on debian stretch, so py35 is the default here.

ThomasWaldmann:
https://docs.python.org/3.5/library/venv.html

also it seems that `venv` would also work for 3.5, so why did you mix raising the python requirement and moving from virtualenv to venv?

ThomasWaldmann:
Also, your PR should be against master now, because we have merged the py3 branch into master already.

Guess it would be best if you first do a PR with everything that works on 3.5 also and delay the 3.6 stuff until later.

eamanu:
Hi @ThomasWaldmann  Thanks for the feedback. 

> Guess it would be best if you first do a PR with everything that works on 3.5 also and delay the 3.6 stuff until later.

Ok I will take a look on the code to help on it. 

> also it seems that venv would also work for 3.5, so why did you mix raising the python requirement and moving from virtualenv to venv?

hmm yes. I think that we can move to venv, this way we avoid the requirement to install other module (virtualenv), because py3 come with venv module.

Issue #890, opened 2019-08-07
refactor charset extraction from contenttype


RogerHaase:
Place this code (or similar) somewhere in utils and refactor various methods of extracting charset to use the new function:

        if 'charset' in rev.meta['contenttype']:
            charset = rev.meta['contenttype'].split('charset=')[1]

There are places where re is used to extract charset. Since presence of charset in contenttype is an indication of a text file, maybe add a helper function to test whether an item is a text-like file.

See #867

ThomasWaldmann:
Maybe the utility function could even go a bit further:

guess one is mostly interested in the charset (coding) because one wants to run `bytes.decode(coding)`.

so the utility function could be just called like:
```
text = as_text(rev)
or even
text = rev.as_text()
or (best)
text = str(rev)
```

RogerHaase:
Try again with `__str__` code added to the Revision object, not the the ProtectedRevision object

RogerHaase:
closing as duplicate of #920

PR #916, opened 2019-08-23
Fixit

RogerHaase:


src/moin/storage/middleware/protecting.py
```diff
@@ -340,6 +340,14 @@ def __init__(self, protector, rev, p_item=None):
         self.rev = rev
         self.item = p_item or ProtectedItem(protector, rev.item)
 
+    def __str__(self):
+        """
+        Return revision contents as a string.
+        """
+        self.data.seek(0)
```
ThomasWaldmann: that looks strange. where does self.data come from?

did you mean `self.rev.data`?

also, this does not really belong into the protecting middleware, so i guess here there should be only a stub that delegates to `rev.__str__` and the implementation should be there.

ThomasWaldmann: also, as this is just another way to get the data, guess we need the same permission checks as in the `data` property, see below.

src/moin/scripts/maint/modify_item.py
```diff
@@ -40,7 +40,7 @@ def run(self, name, meta_file, data_file, revid):
         if 'charset' in rev.meta['contenttype']:
             # input data will have \r\n line endings, output will have platform dependent line endings
             charset = rev.meta['contenttype'].split('charset=')[1]
-            data = rev.data.read().decode(charset)
+            data = str(rev.data)
```
ThomasWaldmann: did you want to call `str(rev)`?

RogerHaase: No that rev is not a ProtectedRevision, so it must be reverted to the old code:

    data = rev.data.read().decode(charset)

src/moin/storage/middleware/protecting.py
RogerHaase: self is a ProtectedRevision object. I think it is best to revert 4024102 and 708ec5d and reopen #890. 

I do not understand what you meant by best = "text = str(rev): in the #890 issue.

Also, seems there is very little code saved.

src/moin/scripts/maint/modify_item.py
ThomasWaldmann: no, the code implementing `__str__` must be in the Revision, the ProtectedRevision then just forwards to that.

src/moin/storage/middleware/protecting.py
ThomasWaldmann: I meant implementing `__str__`.

Just do not put code that has nothing to do with "protection" (== permissions/acl checking) into the protection middleware.

That functionality is rather fundamental and thus should be implemented in the Revision class.
And of course there also needs to be a __str__ in the ProtectedRevision, but that would just check permisssions and then forward to the unprotected Revision.

RogerHaase: ProtectedRevision already has:

    @property
    def meta(self):
        self.require(READ, PUBREAD)
        return self.rev.meta

    @property
    def data(self):
        self.require(READ, PUBREAD)
        return self.rev.data

The above seem to get called when .data is referenced by a call to the `__str__`  method, so no need to check again.

ThomasWaldmann: You would not call these, but `str(self.rev)` - and before doing that, check permissions (only once).

PR #917, opened 2019-08-25
Fixit

RogerHaase:


src/moin/apps/frontend/views.py
```diff
@@ -655,7 +655,8 @@ def convert_item(item_name):
 
     namespace = item.rev.meta.revision.namespace or 'default'
     revno = item.rev.revid
-    content = str(item.rev)
+    item.rev.data.seek(0)
+    content = item.rev.data.read()
```
ThomasWaldmann: i would assume content is bytes now (and originally was str)?!

PR #919, opened 2019-08-26
py3: fix bytes vs strings issues in import19.py and _utils.py

RogerHaase:
todo: non-ascii names get mangled, items processed in random order

### DIFF
--- a/src/moin/scripts/migration/moin19/_utils19.py
+++ b/src/moin/scripts/migration/moin19/_utils19.py
@@ -126,10 +126,6 @@ def unquoteWikiname(filename, charset=CHARSET):
     :rtype: unicode
     :returns: WikiName
     """
-    # From some places we get called with Unicode strings
-    if isinstance(filename, str):
-        filename = filename.encode(CHARSET)
-
     parts = []
     start = 0
     for needle in QUOTED.finditer(filename):
@@ -157,4 +153,4 @@ def unquoteWikiname(filename, charset=CHARSET):
         parts.append(filename[start:len(filename)])
         wikiname = ''.join(parts)
 
-    return wikiname.decode(charset)
+    return wikiname
--- a/src/moin/scripts/migration/moin19/import19.py
+++ b/src/moin/scripts/migration/moin19/import19.py
@@ -204,7 +204,9 @@ class PageBackend:
 
     def __iter__(self):
         pages_dir = os.path.join(self._path, 'pages')
-        for f in os.listdir(pages_dir):
+        pages = os.listdir(pages_dir)
+        pages.sort()
+        for f in pages:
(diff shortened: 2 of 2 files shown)


src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -157,4 +153,4 @@ def unquoteWikiname(filename, charset=CHARSET):
         parts.append(filename[start:len(filename)])
         wikiname = ''.join(parts)
 
-    return wikiname.decode(charset)
```
ThomasWaldmann: guess that code is wrong (before and after your changeset).

if there is (xxyyzz) in a filename, it is a utf-8 sequence - it could be e.g. one 3-byte multibyte-encoded character. xx, yy, zz are not single unicode codepoints, but the code here treats them like that.

i remember vaguely fixing that issue, maybe at some other place in the code?

src/moin/scripts/migration/moin19/import19.py
```diff
@@ -136,7 +136,9 @@ def run(self, data_dir=None):
             if namespace == '':
                 namespace = 'default'
             meta, data = backend.retrieve(namespace, revno)
-            data_in = ''.join(data.readlines())
+            charset = meta['contenttype'].split('charset=')[1]
```
ThomasWaldmann: this code is copy and pasted at quite some places now...

'text/x.moin.wiki;format=1.9;charset=utf-8' <- there you see that this operation is not even working for all the cases (just imagine format and charset were swapped in position).

ThomasWaldmann: i'll open a ticket for this.

src/moin/scripts/migration/moin19/_utils19.py
ThomasWaldmann: https://github.com/moinwiki/moin/blob/master/src/moin/utils/iri.py#L361 there is some similar transformation.

https://github.com/moinwiki/moin/blob/master/src/moin/utils/iri.py#L335 this is the other direction.

src/moin/scripts/migration/moin19/import19.py
```diff
@@ -204,7 +204,9 @@ def __init__(self, path, deleted_mode=DELETED_MODE_KEEP,
 
     def __iter__(self):
         pages_dir = os.path.join(self._path, 'pages')
-        for f in os.listdir(pages_dir):
+        pages = os.listdir(pages_dir)
+        pages.sort()
```
ThomasWaldmann: there is also `sorted` which would make this:
```
pages = sorted(os.listdir(pages_dir))
```

One must be aware that this is sorting the encoded directory names, not the decoded page names, though.

PR #922, opened 2019-09-24
fix #921 import19 fails with py3 bytes vs. strings issue

### LINKED ISSUE #921: import19 fails with py3 bytes vs. strings issue
Converting Pages/Attachments...

      File "c:\GIT\moin-venv-python\Scripts\moin-script.py", line 11, in <module>
        load_entry_point('moin', 'console_scripts', 'moin')()
      File "c:\git\moin\src\moin\scripts\__init__.py", line 59, in main
        return manager.run(default_command=default_command)
      File "c:\git\moin-venv-python\lib\site-packages\flask_script\__init__.py", line 417, in run
        result = self.handle(argv[0], argv[1:])
      File "c:\git\moin-venv-python\lib\site-packages\flask_script\__init__.py", line 386, in handle
        res = handle(*args, **config)
      File "c:\git\moin-venv-python\lib\site-packages\flask_script\commands.py", line 216, in __call__
        return self.run(*args, **kwargs)
      File "c:\git\moin\src\moin\scripts\migration\moin19\import19.py", line 105, in run
        for rev in PageBackend(data_dir, deleted_mode=DELETED_MODE_KILL, default_markup='wiki'):
      File "c:\git\moin\src\moin\scripts\migration\moin19\import19.py", line 206, in __iter__
        itemname = unquoteWikiname(f)
      File "c:\git\moin\src\moin\scripts\migration\moin19\_utils19.py", line 135, in unquoteWikiname
        for needle in QUOTED.finditer(filename):
    TypeError: <moin.scripts.migration.moin19.import19.ImportMoin19 object at 0x0000029A13F49A08>: cannot use a string pattern on a bytes-like object

RogerHaase:


src/moin/scripts/migration/moin19/import19.py
```diff
@@ -65,10 +63,12 @@
     'creole': 'text/x.moin.creole;charset=utf-8',
     'python': 'text/x-python;charset=utf-8',
     'text/creole': 'text/x.moin.creole;charset=utf-8',
-    'rst': 'text/rst;charset=utf-8',
-    'text/rst': 'text/rst;charset=utf-8',
+    'rst': 'text/x-rst;charset=utf-8',
+    'text/rst': 'text/x-rst;charset=utf-8',
```
ThomasWaldmann: hmm, correct.

strange, wondering how we found the wrong mimetype...

src/moin/scripts/migration/moin19/import19.py
```diff
@@ -335,8 +337,12 @@ def __init__(self, item, revno, path):
             meta, data = split_body(content)
         meta.update(editlog_data)
         format = meta.pop('format', self.backend.format_default)
+        if format.startswith('csv'):
+            format = 'csv'  # drop trailing sep character as in "format csv ;"
```
ThomasWaldmann: why is this? why is it only for csv?

src/moin/converters/text_csv_in.py
```diff
@@ -28,7 +28,7 @@ def _factory(cls, type_input, type_output, **kw):
     def __call__(self, data, contenttype=None, arguments=None):
         text = decode_data(data, contenttype)
         content = normalize_split_text(text)
-        dialect = csv.Sniffer().sniff(content[0])
+        dialect = csv.Sniffer().sniff(text)
```
ThomasWaldmann: so, the sniffer needs more than the first line AND is happy no matter what the line ends are?

src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -8,8 +8,8 @@
 import re
 
 from moin.constants.keys import NAME, ACL, CONTENTTYPE, MTIME, LANGUAGE
+from moin.constants.contenttypes import CHARSET
```
ThomasWaldmann: well, maybe nitpicking, but guess that should be separated, like CHARSET19.

imagine we ever change the CHARSET, the CHARSET19 should stay same, right?

src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -126,15 +126,11 @@ def unquoteWikiname(filename, charset=CHARSET):
     :rtype: unicode
     :returns: WikiName
     """
-    # From some places we get called with Unicode strings
-    if isinstance(filename, str):
-        filename = filename.encode(CHARSET)
```
ThomasWaldmann: if this normalization is removed, shouldn't there be at least a type assertion here, so we can be sure with what we deal here?

src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -126,15 +126,11 @@ def unquoteWikiname(filename, charset=CHARSET):
...
-
     parts = []
     start = 0
     for needle in QUOTED.finditer(filename):
         # append leading unquoted stuff
-        parts.append(filename[start:needle.start()])
+        parts.append(bytes(filename[start:needle.start()].encode(charset)))
```
ThomasWaldmann: this looks suspicious.

type of filename? guess this should be str, otherwise .encode() would not make sense.

after .encode(), we have bytes, so why another bytes() around that?

src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -144,17 +140,17 @@ def unquoteWikiname(filename, charset=CHARSET):
...
-                parts.append(character)
+                parts.append(bytes.fromhex(byte))
         except ValueError:
             # byte not in hex, e.g 'xy'
             raise InvalidFileNameError(filename)
 
     # append rest of string
     if start == 0:
-        wikiname = filename
+        wikiname = bytes(filename.encode(charset))
```
ThomasWaldmann: isn't .encode() already returning bytes type?

src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -144,17 +140,17 @@ def unquoteWikiname(filename, charset=CHARSET):
...
     else:
-        parts.append(filename[start:len(filename)])
-        wikiname = ''.join(parts)
+        parts.append(bytes(filename[start:len(filename)].encode(charset)))
+        wikiname = b''.join(parts)
 
-    return wikiname.decode(charset)
+    wikiname = wikiname.decode(charset)
+    return wikiname
```
ThomasWaldmann: don't think this change is needed.

src/moin/scripts/migration/moin19/_utils19.py
```diff
@@ -144,17 +140,17 @@ def unquoteWikiname(filename, charset=CHARSET):
...
             raise InvalidFileNameError(filename)
 
     # append rest of string
     if start == 0:
-        wikiname = filename
+        wikiname = bytes(filename.encode(charset))
     else:
-        parts.append(filename[start:len(filename)])
-        wikiname = ''.join(parts)
+        parts.append(bytes(filename[start:len(filename)].encode(charset)))
```
ThomasWaldmann: same here.

src/moin/scripts/migration/moin19/import19.py
RogerHaase: Probably never tested, or not tested recently

src/moin/scripts/migration/moin19/import19.py
RogerHaase: the next line is:

meta[CONTENTTYPE] = FORMAT_TO_CONTENTTYPE.get(format, CONTENTTYPE_DEFAULT)

csv is recognized, but "csv ,"  defaults to plain text.

I am not aware of other #format commands that pass a parameter similar to "#format csv ,"

src/moin/converters/text_csv_in.py
RogerHaase: in my testing, sniffer objects to leading blank lines and input passed as a list.

I have not tested on Ubuntu

src/moin/scripts/migration/moin19/_utils19.py
RogerHaase: fixed in next patch

src/moin/scripts/migration/moin19/_utils19.py
RogerHaase: unquoteWikiname method was copied from wikiutil.py in 1.9 where is was called from several places. Here it is called from only 2 places within import19.py and both places pass string(unicode).

src/moin/scripts/migration/moin19/_utils19.py
RogerHaase: will remove redundant bytes()

src/moin/scripts/migration/moin19/_utils19.py
RogerHaase: yes

src/moin/scripts/migration/moin19/_utils19.py
RogerHaase: will fix in next patch

src/moin/scripts/migration/moin19/_utils19.py
RogerHaase: will fix in next patch

src/moin/converters/text_csv_in.py
ThomasWaldmann: ok, let's try it.

src/moin/scripts/migration/moin19/_utils19.py
ThomasWaldmann: ok

src/moin/scripts/migration/moin19/import19.py
ThomasWaldmann: ok. if this was a general mechanism, we should split at ";" and then take result[0].strip().

ThomasWaldmann:
Last unclear thing:

https://github.com/moinwiki/moin/pull/922#commitcomment-35271017

Guess it is ready to merge after that.

Issue #925, opened 2019-11-07
Support single quotes in extension (macro) arguments


lyind:
Sometimes users try to use macros to render text with quotes and commas included, like:

`<<Color2({"comment": "bla","user":"id","date":"DATETIME"},green)>>`

Since the comma is used as argument separator, rendering will not work as intended. Quoting the string is not possible since `"` occurs in the text.

The character `"` is the only quote character specified in `wikiutil.py:parse_quoted_separated_text()`.

I think it would be helpful for users to extend the list of permitted quote characters to include `'` (single quote). The code already provides the logic to handle this.

The example above could then be written as:
`<<Color2('{"comment": "bla","user":"id","date":"DATETIME"}',green)>>`

This will pass the correct arguments to the extension and work as expected from the users point of view.

This change would break formatting of existing wiki text passing arguments starting with a single-quote (`'`) as part of the actual argument. I consider the occurence of such arguments highly unlikely, though.

ThomasWaldmann:
> Quoting the string is not possible since " occurs in the text.

Well, i guess it should be possible like:
```
<<Color2("{""comment"": ""bla"",""user"":""id"",""date"":""DATETIME""}",green)>>
```

But this is not pretty, so enabling single quote also sounds like a good idea.

@RogerHaase what do you think?

RogerHaase:
I am recovering from cataract surgery so I can barely read this text and
will not get new glasses until January.

Assuming Color2 is a custom macro, have you tried just passing the the
quotes parameter to parse_quoted_separated_ext? If that doesn't work the
change is OK.

lyind:
@RogerHaase Thank you for you great work and get well soon!

@ThomasWaldmann Can you please assist me on how to accomplish what @RogerHaase suggested?

1. I don't know if [Color2](http://moinmo.in/MacroMarket/Color2) is classified as custom macro. What is the difference between a custom and a non-custom (builtin?) macro?
2. I also don't know how to move the call to `parse_quoted_separated_ext()` inside the macro as the function is normally called in wikiutil's `invoke_extension_function()` for all extension arguments.

I realize a reduced scope of changes limits potential risk. However, I think supporting `'` for quoting macro arguments indeed makes sense for all macros with arguments.

ThomasWaldmann:
@RogerHaase all the best for you and get well soon!

@lyind if it is in the moin repo, it is builtin, if not, it is 3rd party / custom.

[...]

PR #926, opened 2019-11-07
[#925] parse_quoted_separated_text(): add single quote to default quotes

lyind:
Implement issue #925.

No additional tests needed as the logic is already tested in `testExtendedParserQuoting()`.


RogerHaase:
sorry for delay, thanks for helping

Issue #720, opened 2018-10-12
port moin2 code to py3


ThomasWaldmann:
See there: https://github.com/moinwiki/moin/projects/1

ThomasWaldmann:
If you want to help with this, something that can be done at any time is to improve testing:
- better test coverage
- better tests
- see also other testing-related tickets

Before we can port to py3 at all, we need to have good testing, otherwise we might introduce lots of issues and manual testing would slow down porting also.

ThomasWaldmann:
I talked to some people at europython 2019 and changed my mind.

We rather port to py3 now:
- people can write new code in py3, which is more enjoyable and what modern developers might expect
- project looks less "dead"
- port to py3 **only** (not py2 + py3)

I suggest to require py35+ for now (as debian stable still had that before buster was released a few days ago) and later increase to py36+

ReimarBauer:
We had some issues with our mss software with py35 and we are now based on py37. That does not mean that these problems will occure in moin too. I only remember that this was a problem on some of the platforms we support. 

We did the same step this year before gsoc starts and that made a big difference for all who want to contribute.

Thx for these good news.

ThomasWaldmann:
We are going to py35+ only, no py2 compat needed, so we do not want to use `six`.

TODO: Fork `ThomasWaldmann:moin`, start with branch py3, create feature branch, pick one todo from below, carefully globally fix all related to it, commit like `py3: dict`, PR against py3 branch.

- [ ] 2to3 dict
- [ ] 2to3 xrange

- [ ] 2to3 imports
- [ ] 2to3 urllib

- [ ] python-modernize print

- [ ] 2to3 basestring
- [ ] 2to3 unicode
- [ ] python-modernize unichr: uses six.unichr

ThomasWaldmann:
easy task (pep8 fixing): https://github.com/moinwiki/moin/pull/845#issuecomment-511183252

ThomasWaldmann:
another task: https://github.com/moinwiki/moin/pull/845#issuecomment-511192283

I'll be away for a while, trying an ebike in basel area, so this is the end to ep2019 sprints for me now. :)

Any work on these recent issues (also see comment above) would be appreciated.

Please base your work on my PR code: https://github.com/moinwiki/moin/pull/845

ThomasWaldmann:
Done and merged into master.

So moin2 is now a python3-only project. \o/

PR #927, opened 2019-12-23
Fixed print syntax for py3

ingoha:
Includes a note for the venv path; needs to be addressed...

### DIFF
--- a/contrib/wsgi/moin2.wsgi
+++ b/contrib/wsgi/moin2.wsgi
@@ -33,15 +33,15 @@ if not (moin_dir in sys.path or moin_dir.lower() in sys.path):
     sys.path.insert(0, moin_dir)
 
 # for debugging sys.path issues, comment out after things are working
-print '== moin2.wsgi sys.path =='
+print('== moin2.wsgi sys.path ==')
 for p in sys.path:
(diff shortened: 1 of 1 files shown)


ingoha:
Amends https://github.com/moinwiki/moin/issues/720

contrib/wsgi/moin2.wsgi
```diff
@@ -26,22 +26,23 @@ moin_dir = '/path/to/moin/dir/where/wikiconfig.py/is/located'
 if sys.platform == 'win32':
     site.addsitedir(moin_dir + '-venv-python/Lib/site-packages')
 else:
+# TODO: the venv is created in ../moin-venv-python3.5 by quickinstall...
```
ThomasWaldmann: so, this is a bug because there is "3.5" missing after "python"?

if so, please do a separate commit or a bug report, it doesn't fit into a "fix the prints" commit.

ingoha: You are right, sorry - moved this issue to https://github.com/moinwiki/moin/issues/929. The comment will be removed from the PR...

ThomasWaldmann:
Thanks!

PR #931, opened 2019-12-25
Properly choose last revision of pages. Fixes #928

### LINKED ISSUE #928: import19 chooses wrong "last revision of Moin 1.9" for markup conversion
In my setup (importing from 1.9.9), the importer chooses arbitrary revisions but not the latest one.
Perhaps this has to do with the usage of revision ids for finding the latest one vs. the revision no (which would be the correct way?)?

ingoha:


RogerHaase:
Thanks for reporting bug and attempting a solution. Alternate fix applied: 
3b627e7

PR #935, opened 2019-12-31
Save Draft creates extra EOLs on windows, fixes #934

### LINKED ISSUE #934: Save Draft creates extra EOLs on windows
If a new item is created that has 2 blank lines between the A and B:

    A
    
    
    B

then a hex view of the file saved in wiki/data/default/data/ looks like (spaces inserted for clarity):

    41 0D0A 0D0A 0D0A 42

But if the Preview button is clicked before the item is saved, then the saved file in wiki/preview/
looks like:

    41 0D 0D0A 0D 0D0A 0D 0D0A 42

e.g. an extra 0D character is added before each 0D0A line end present in the textarea. 

This problem happens only on windows, and is caused by this code:

        if data_in:
            with open(self.draft_name, 'w', encoding=self.coding) as f:
                f.write(data_in)

RogerHaase:


src/moin/utils/edit_locking.py
```diff
@@ -159,7 +159,8 @@ def put_draft(self, data_in, overwrite=True):
                 self.draft_name = self.make_draft_name(rev_id)
             self.cursor.execute('''DELETE FROM editdraft WHERE user_name = ? ''', (self.user_name, ))
         if data_in:
-            with open(self.draft_name, 'w', encoding=self.coding) as f:
+            data_in = data_in.encode(self.coding)
+            with open(self.draft_name, 'wb') as f:
```
ThomasWaldmann: when using "wb" here, maybe also use "rb" below?

PR #938, opened 2020-01-06
import19 chooses wrong "last revision of Moin 1.9" for markup convers…

### LINKED ISSUE #928: import19 chooses wrong "last revision of Moin 1.9" for markup conversion

RogerHaase:
…ion fixes #928

### DIFF
--- a/src/moin/scripts/migration/moin19/import19.py
+++ b/src/moin/scripts/migration/moin19/import19.py
@@ -245,8 +245,8 @@ class PageItem:
     def iter_revisions(self):
         revisionspath = os.path.join(self.path, 'revisions')
         try:
-            # rather use this or a range(1, self.current+1)?
-            fnames = os.listdir(revisionspath)
+            # alternative method is to generate file names using range(1, self.current+1)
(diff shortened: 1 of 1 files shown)


src/moin/scripts/migration/moin19/import19.py
```diff
@@ -245,8 +245,8 @@ def __init__(self, backend, path, itemname):
     def iter_revisions(self):
         revisionspath = os.path.join(self.path, 'revisions')
         try:
-            # rather use this or a range(1, self.current+1)?
-            fnames = os.listdir(revisionspath)
+            # alternative method is to generate file names using range(1, self.current+1)
+            fnames = sorted(os.listdir(revisionspath))
```
ThomasWaldmann: good that "back then", i formatted the rev file names with leading zeros, so sorted() now just works. :-)

PR #945, opened 2020-01-19
improve performance of item history, fixes #940

### LINKED ISSUE #940: Item History is sluggish
After a couple of timers to the history procedure of frontend/views.py near line 1316:

    flaskg.clock.start('runrevs')
    for rev in revs:
        entry = dict(rev.meta)
        entry[FQNAME] = rev.fqname
        entry[FQNAMES] = rev.fqnames
        history.append(entry)
        close_file(rev.data)
    flaskg.clock.stop('runrevs')

    history_page = utils.getPageContent(history, offset, results_per_page)
    close_file(item.rev.data)
    trash = item.meta['trash'] if 'trash' in item.meta else False
    
    flaskg.clock.start('renderrevs')
    ret = render_template('history.html',
                          fqname=fqname,
                          item=item,
                          item_name=item_name,
                          history_page=history_page,
                          bookmark_time=bookmark_time,
                          NAME_EXACT=NAME_EXACT,
                          len=len,
                          trash=trash,
    )
    flaskg.clock.stop('renderrevs')
    
    close_file(item.rev.data)
    return ret

and then displaying 20 revisions of an item with many revisions:

    runrevs takes 39682.98ms
    renderrevs takes 5204.00ms

Displaying the second page of revisions yields similar timings.

One reason runrevs is taking a long time is it appears the content of each revision is needlessly read; only a few fields of the metadata are required.  

Is there a way to read only the metadata of a revision?

RogerHaase:


src/moin/storage/middleware/indexing.py
```diff
@@ -792,6 +792,13 @@ def search_page(self, q, idx_name=LATEST_REVS, pagenum=1, pagelen=10, **kw):
                 item = Item(self, latest_doc=latest_doc, itemid=doc[ITEMID])
                 yield item.get_revision(doc[REVID], doc=doc)
 
+    def search_len(self, q, idx_name=ALL_REVS, **kw):
+        """
+        Return the number of matching revisions.
+        """
+        with self.ix[idx_name].searcher() as searcher:
+            return len(searcher.search(q, **kw))
```
ThomasWaldmann: this seems somehow expensive/wasteful to run the search and only use the length of the result, but not the result itself.

ThomasWaldmann: also, the method name ... - maybe rather `result_size`?

Issue #946, opened 2020-01-21
Global Index, History, Tags are sluggish


RogerHaase:
Takes over 100 seconds to view the global index on a medium sized wiki, where _latest_revs_1.toc is 34,000  bytes and latest_revs_......seg is 38,000,000 bytes.

RogerHaase:
Almost all the processing time is within the make_flat_index method in `items/__init__.py`. In a recent test, 133.5 seconds elapsed in the for loop running through the revisions, total time was 135.0 seconds.

I think the only way to speed this up is to try making a skinny_latest_revs index that contains only the metadata required for the Global Index, Global History (186 seconds with no bookmark set; 38 seconds with bookmark set showing less than 150 items), Global Tags (94  seconds to find 1 tag), user/wanted-items (133  seconds to find 233 items), user/orphaned-items (103 seconds to find 757 items).

RogerHaase:
A hack to try a skinny index, yields a Global Index in less than 5 seconds. But lots of stuff to fix: ACL checking, updating index when an item is deleted, modified, etc.

RogerHaase:
Some sample timings for sluggish transactions on a wiki imported from moin 1.9:
 * unique default items: 1100
 * top level default items: 160
 * default revisions: 8300
 * user profiles: 50
 * user revisions: 1300
 * wiki/data/default/data size: 155mb

Timings below are in seconds, running built in server on Windows 10. Other activities were sometimes performed while waiting for long running transactions.

|Transaction|hierarchic=True|hierarchic=False|comment|
|------------------|-------------:|-----:|:-------|
|global index|98|69|  |
|global history|187|170 ||
|tags|64|55||
|admin item ACL report|244|269|odd that hierarchic=True is faster|
|admin user ACL info|437|421||
|user wanted items|72|67||
|user orphaned items|73|66||
|user item sizes|72|65||
|user my changes - <10 changes |4|4||
|user my changes - > 5000 changes|584|665| showing first 50 changes|
|user my changes - > 5000 changes|484|521| showing second 50 changes|
|item history|7|7|first 50 revisions; code recently improved #940|

RogerHaase:
As an alternative to adding a new index, key-latest-revs, tried removing the content from latest-revs and all-revs. This reduced the response times for Global Index and Global History by about 20%. File sizes are shown below. Conclusion is that the size of the ...revs...seg file is a major factor when preparing a report that includes all current revs.

[...]

PR #953, opened 2020-02-09
reduce response time for global index by about 90% #946

RogerHaase:


ThomasWaldmann:
I am missing a good commit comment that describes what the problem was, what has been found to be the root cause of the problem and how it was solved.

RogerHaase:
Will postpone adding a small index until other methods are exhausted. Use of a small index resulted in fastest response times, but maybe use of the latest-revs index can be made fast enough.

PR #970, opened 2020-03-11
avoid tracebacks in import19 caused by lack of unicode support in Win…

RogerHaase:
…dows

### DIFF
--- a/src/moin/scripts/migration/moin19/import19.py
+++ b/src/moin/scripts/migration/moin19/import19.py
@@ -131,7 +131,11 @@ class ImportMoin19(Command):
         reg = default_registry
         refs_conv = reg.get(type_moin_document, type_moin_document, items='refs')
         for item_name, (revno, namespace) in sorted(last_moin19_rev.items()):
-            print('    Processing item "{0}", namespace "{1}", revision "{2}"'.format(item_name, namespace, revno))
+            try:
+                print('    Processing item "{0}", namespace "{1}", revision "{2}"'.format(item_name, namespace, revno))
+            except UnicodeEncodeError:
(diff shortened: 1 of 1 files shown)


src/moin/scripts/migration/moin19/import19.py
```diff
@@ -131,7 +131,11 @@ def run(self, data_dir=None):
         reg = default_registry
         refs_conv = reg.get(type_moin_document, type_moin_document, items='refs')
         for item_name, (revno, namespace) in sorted(last_moin19_rev.items()):
-            print('    Processing item "{0}", namespace "{1}", revision "{2}"'.format(item_name, namespace, revno))
+            try:
+                print('    Processing item "{0}", namespace "{1}", revision "{2}"'.format(item_name, namespace, revno))
+            except UnicodeEncodeError:
+                print('    Processing item "{0}", namespace "{1}", revision "{2}"'.format(
+                      'with unicode name Windows cannot display', namespace, revno))
```
ThomasWaldmann: Huh? Windows not supporting unicode? I thought it is based on ucs2 (16bit unicode) since ages?

And even if, there is .encode('ascii', errors='replace') or so.

RogerHaase: I copied  AjudaNaFormatação, StronaGłówna, and ШаблонДомашняяСтраница from the moin2 sample wiki to a 1.9 test wiki and then ran "m import19 ...". AjudaNaFormatação prints OK, the other 2 fail with UnicodeEncodeError. 

Perhaps a complicating factor is that printed output going to sys.stdout and sys.stderr are being redirected to a file.

Will change the print after `except UnicodeEncodeError:` to include `item_name.encode('ascii', errors='replace')` and remove 'with unicode name Windows cannot display'.

PR #991, opened 2020-04-03
BF+ENH: Error msg instead of 500 on item conversion failure

moloney:
Also show info message on successful conversion.


RogerHaase:
thanks for helping.

Issue #909, opened 2019-08-17
remove ./m and quickinstall references...


ThomasWaldmann:
... from everywhere except development docs.

ThomasWaldmann:
also virtualenv.

Issue #908, opened 2019-08-17
docs say git and virtualenv are required


ThomasWaldmann:
https://moin-20.readthedocs.io/en/latest/admin/requirements.html#requirements

remove this from there, this is stuff for developers and we should not bother normal users/admins with it.

Issue #907, opened 2019-08-17
create user install docs


ThomasWaldmann:
- install via pip, apt, ...
- `moin` command based steps to get a wiki working (do not use quickinstall.py or "m", users do not have this)

ThomasWaldmann:
in general, no docs should refer to "m" except the developer install section.

RogerHaase:
I believe https://moin-20.readthedocs.io/en/latest/admin/install.html#installation solves this issue. Closing.

Issue #906, opened 2019-08-17
move dev docs to "development" chapter


ThomasWaldmann:
https://moin-20.readthedocs.io/en/latest/admin/install.html

this is all based on git clone, quickinstall.py and "m" - developers only, normal users do not have this.

https://moin-20.readthedocs.io/en/latest/man/moin.html

again, normal users do not have "m".

PR #993, opened 2020-04-06
ENH: Add 'docs' extra to setup.py to install deps for doc building

moloney:
Allows "pip install moin[docs]" to get documentation build deps.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,9 @@ setup_args = dict(
                                          # requires special libs/header to be installed before it can be compiled successfully
                                          # windows binaries available from 3rd parties
         'sqla': ["sqlalchemy>=0.7.1"],  # used by sqla store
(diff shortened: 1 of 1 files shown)


RogerHaase:
This change is not required because if you do ./m it will say:

    extras          install packages required for docs and moin development

doing ./m extras will install the packages in requirements.d/docs.txt: sphinx and sphinx_rtd_theme. And also the packages in requirement.d/development.txt.

The install process could be made better, there is #906, #907, #908, #909...

moloney:
I don't have strong feelings on this, but looking through Issues/PRs I got the impression that the project is trying to move towards a more standard python workflow.  The standard way to install extras in python is doing stuff like "pip install moin[docs]", and the other "extras" are already in setup.py so I figured I would add the last missing piece.

PR #1011, opened 2020-05-09
fix several deprecationWarnings

sblondon:
This PR fixes several `DeprecationWarning` shown when `tox` runs. They are fixed by updating the `import`.

There are still two DeprecationWarnings which can not be fixed this way: one DeprecationWarning is in `flask_theme` dependancy and the other one is about `werkzeug.contrib.atom` module which will be removed in Werkzeug 1.0.

### DIFF
--- a/src/moin/auth/__init__.py
+++ b/src/moin/auth/__init__.py
@@ -133,7 +133,9 @@
     name of the authentication method.
 """
 
-from werkzeug import redirect, abort, url_quote, url_quote_plus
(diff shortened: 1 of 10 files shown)


sblondon:
I sent a PR to Flask-theme (https://github.com/liuliqiang/flask-theme/pull/11) to fix the `DeprecationWarning` in this library.

RogerHaase:
Thanks for helping!

Issue #850, opened 2019-07-19
atom code deprecation?


ThomasWaldmann:
```
/home/user/w/moin/src/moin/apps/feed/views.py:18:
DeprecationWarning: 'werkzeug.contrib.atom' is deprecated as of version 0.15 and will be removed in version 1.0.
```

Strange, back then I had the impression that atom is the more modern/cleaner approach than rss.

Does anyone know why?

What's the current "state of art" regarding feeds?

ThomasWaldmann:
https://github.com/lkiesow/python-feedgen/ maybe we could use this.

the werkzeug guys recommend using some feed generation library (but don't point to anything specific).

sblondon:
According to the [Changelog](https://werkzeug.palletsprojects.com/en/1.0.x/changes/#version-0-16-1), the removal is not about Atom vs RSS: it 'did not fit in with the rest of Werkzeug, and is better served by a dedicated library in the community'.

python-feedgen seems to fit well (python3, available at pypi.org). I can write a PR if you're interested.

RogerHaase:
Please do a PR, thanks!

PR #1013, opened 2020-05-17
generate Atom feed with feedgen library instead of Flask

sblondon:
This PR uses feedgen library to generation Atom feed.

I checked the tests pass with `tox --recreate` for python 3.7 . The previous output (generated with `werkzeug.contrib.atom.AtomFeed()`) is more or less the same with feedgen output in feed test.
 
I failed to run moinmoin so I did not see the real output.

Closes: #850


RogerHaase:
Thanks for helping!

Issue #1015, opened 2020-05-19
Atom feed needs more testing and maybe fixing


RogerHaase:
This problem occurred prior to conversion to feedgen library and persists after the change.

Using Freeder on an Android phone, the message "MoinMoin feels unhappy"  frequently appears when accessing 

    <moin2 wiki url>/+feed/atom or <moin2 wiki url>/+feed/atom/Home

When given the same url, Chrome and Opera display the raw XML file, Firefox presents a download dialog.

The problem is in /apps/feed/views.py:

            try:
                hl_item = Item.create(name, rev_id=this_revid)
                if previous_revid is not None:
                    # HTML diff for subsequent revisions
                    previous_rev = item[previous_revid]
                    content = hl_item.content._render_data_diff_atom(previous_rev, this_rev)
                else:
                    # full html rendering for new items
                    content = render_template('atom.html', get='first_revision', rev=this_rev,
                                              content=Markup(hl_item.content._render_data()), revision=this_revid)
                content_type = 'html'
            except Exception as e:
                logging.exception("content rendering crashed")
                content = _('MoinMoin feels unhappy.')
                content_type = 'text'

sblondon:
The try/except block did not change during the migration from `flask.atom` to `feedgen` library. So the bug is still there.

I have two hypothesis : 
- Freeder does not use the correct URL (for example by adding parameters which are badly interpreted)
- the code is called if 'content is None' https://github.com/moinwiki/moin/blob/9af32d3cd56b32792bb804df081add97e05b1b84/src/moin/apps/feed/views.py#L55 This occurs if `revs` is an empty list, which is about searching in cached modifications if I understand properly. Perhaps you see the error with Freeder because it downloaded several times along the day the modifications so sometimes there are no modification) than a browser. Did you try with/without cache/private mode in browser? Or with `wget` two times  in interval of few seconds?

RogerHaase:
My initial report was unclear. The MoinMoin is unhappy message is also in the raw XML files and the downloaded file.

sblondon:
Adding info in the logging step (probably via `stack_info` or `exc_info`) could help to understand the bug.

Are you interested by a PR?

RogerHaase:
yes

RogerHaase:
Does the above imply that you cannot reproduce the error? On a freshly loaded sample wiki with one small change I get:

`
  <entry>
[...]

PR #1016, opened 2020-05-24
fix Atom feed when feed includes revision

sblondon:
When the Atom feed includes diff with a previous revision, an exception is raised because `fqname` 
is not defined (see https://github.com/moinwiki/moin/issues/1015).
This patch adds the `fqname` parameter to `Text._render_data_diff_atom()` method (like `Text._render_data_diff_html()` and `Text._render_data_diff()` methods) and uses it.

Closes: https://github.com/moinwiki/moin/issues/1015


RogerHaase:
Thanks for helping. 

Discovered #1017 while testing your fix.  Will push fix for it shortly.

Issue #951, opened 2020-02-08
werkzeug 1.0.0 cannot import cached_property


RogerHaase:
If quickinstall installs werkzeug 1.0.0, you must install older version:

    pip install werkzeug==0.16.1

sblondon:
According to a warning displayed when `tox` runs, the `import` is done in flask-theme:

```
.tox/py37/lib/python3.7/site-packages/flask_theme/__init__.py:27
  /home/stephane/src/moin/.tox/py37/lib/python3.7/site-packages/flask_theme/__init__.py:27: DeprecationWarning: The import 'werkzeug.cached_property' is deprecated and will be removed in Werkzeug 1.0. Use 'from werkzeug.utils import cached_property' instead.
    from werkzeug import cached_property
```

It's fixed in [upstream repository](https://github.com/liuliqiang/flask-theme). When a new release will be published in pypi.org, this error will disappear.
Updating the dependancy version from flask-theme should be the right fix.

PR #1028, opened 2020-06-02
Jinja and Werkzeug dependancies are included in Flask

sblondon:
Flask already depends from Jinja and Werkzeug.

As the constraints about Jinja and Werkzeug allow latest versions, they are unecessary. So it's probably better to depend only on Flask.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,9 +86,7 @@ setup_args = dict(
         'emeraldtree>=0.10.0',  # xml processing
         'feedgen==0.9.*',  # Atom feed
         'flatland>=0.8',  # form handling
-        'Jinja2>=2.7',  # template engine
         'pygments>=1.4',  # src code / text file highlighting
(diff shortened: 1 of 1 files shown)


RogerHaase:
With #951 setup.py had a workaround where werkzeug was nailed at version <1.0.0 for several months. If we remove the entries for jinja and werkzeug, do we lose the ability to specify a release where it is clearly visible? New werkzeug releases have frequently caused trouble.

sblondon:
The current dependency on Flask is `Flask>=0.10` and the last released version on pypi.org is 1.1.2. I think you should control more the used Flask library than trying to manage the dependancies of Flask.

For example, I can add `Flask==1.1.*` to this PR.

If you're not convinced, just close it. I use Werkzeug and Jinja at work for several years but I'm a beginer with Flask so I don't have experience how Flask manage well (or not) their own dependancies.

RogerHaase:
closing per above, thanks for helping.

Issue #1052, opened 2020-07-23
m css yields: (node:25556) Warning: Accessing non-existent property 'lineno'


RogerHaase:
Upgrading Node to version 14 results in following harmless warning messages:

        (moin-venv-python) C:\git\moin>m css
        Running Stylus to update Modernized theme CSS files...
        (node:25556) Warning: Accessing non-existent property 'lineno' of module exports inside circular dependency
        (Use `node --trace-warnings ...` to show where the warning was created)
        (node:25556) Warning: Accessing non-existent property 'filename' of module exports inside circular dependency
        (node:25556) Warning: Accessing non-existent property 'lineno' of module exports inside circular dependency
        (node:25556) Warning: Accessing non-existent property 'filename' of module exports inside circular dependency
        Success: Modernized CSS files updated.

        Starting directory is C:\git\moin\src\moin\themes\modernized\static\css
        Running lessc to update Basic theme CSS files...
        Success: Basic theme CSS files updated.

The modernized css file is created correctly.

Node https://github.com/nodejs/node/issues/32987 blames it on Stylus https://github.com/stylus/stylus/issues/2534.

Stylus says The issue was fixed by #2538. Current stylus 0.54.8 does not display these warnings anymore.

But we use Stylus 42 per docs:

    sudo npm install stylus@0.42.2 -g  # Windows: npm install stylus@0.42.2 -g
    # we need 0.42.2 because with more recent versions, --compress compresses
    # complete output to 1 line (0.42.2 compresses to 1 line per rule)
    stylus -V  # show version number to prove it works

Stylus has been PITA, best alternative is to change Modernized theme to use CSS Variables.

Alternatives:
* use Node version 13.
* ignore the harmless messages

PR #1059, opened 2020-08-06
Stylus

RogerHaase:


ThomasWaldmann:
first commit: looks ok (just does not really belong into scope of "Stylus" PR).

src/moin/templates/base.html
```diff
@@ -53,10 +53,9 @@
 
                 {%- block theme_stylesheets %}
                     <link media="all" rel="stylesheet" href="{{ url_for('static', filename='css/normalize.css') }}" />
+                    <link media="all" rel="stylesheet" href="{{ url_for('static', filename='css/variables.css') }}" />
                     {% if theme_supp.variables_css() %}
                         <link media="all" rel="stylesheet" href="{{ theme_static('css/variables.css') }}" />
-                    {% else %}
-                        <link media="all" rel="stylesheet" href="{{ url_for('static', filename='css/variables.css') }}" />
```
ThomasWaldmann, line 59: guess the idea of the old code was that a theme's variables.css may or may not @import the default variables.css.

ThomasWaldmann:
is there an issue which this PR is fixing or why are we removing stylus?

src/moin/templates/base.html
RogerHaase: My understanding is browsers load CSS files in parallel, but @import files cannot be loaded until CSS parsing begins. Small benefit.

Old code loaded only 1 variables.css assuming that if a theme had a variables.css it was going to override all the common.css colors; maybe implement a dark theme. New code allows a theme to have a variables.css if it wants one.

RogerHaase:
> is there an issue which this PR is fixing or why are we removing stylus?

#1052 

CSS variables seem easier to use than preprocessors in general. Also more powerful should you choose to modify the variables with Javascript. The only downside to use of CSS variables is IE11 market share.

Long ago I thought I would like using CSS preprocessors, but I now feel they are all time wasters and CSS variables will eventually rule.

ThomasWaldmann:
OK, if we can live without that, feel free to remove. IIRC another nice thing in stylus was python-like syntax. Guess we can't have that with css directly. But if there is too much pain with stylus elsewhere, we can drop it.

I didn't review the css changes in here due to their size.

Issue #700, opened 2018-09-30
Rewrite and document edit display of item meta data


RogerHaase:
When an item is edited, the item's meta data is displayed and may be changed by the user. There are no docs describing how to change these fields. There are other open issues related to meta data.

ideas:

* display extra metadata in a format that can not be edited
  * if some "extra" items (namespace) can be changed, then move to a new heading/input area
    * show namespaces choices
* move item type to extra metadata until blogs or tickets are reintroduced to ui
  * is it possible to make a default item into a ticket/ blog?
* add hints indicating whether comma separated or blank separated is supported on tags & names
* a help link for content type
* if user is allowed to change Item Type and Content Type in metadata, then update meta_changed method in `/items/__init__.py`

Issue #580, opened 2018-03-28
Item Views Rename and Item Views Delete should support alias names like Item Modify


ThomasWaldmann:
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/580) by RogerHaase (Bitbucket: [RogerHaase](https://bitbucket.org/RogerHaase), GitHub: [RogerHaase](https://github.com/RogerHaase)).**

----------------------------------------

If you modify an item, the General Meta Names field shows all item names as a comma separated string. A user may change any of the names as well as add or delete names.

But if a user clicks the Item View Rename link, only the current used name is shown. If the user tries to enter a comma separated list of names, then the resulting new name will include the entire string with the comma characters. 

This situation is corrupted further by doing a Modify and Save. The names field is filled with the erroneous name and these are recognized as multiple names on save. However the longer name with embedded commas and spaces persists.

The Item View Delete dialog has a similar issue. The delete dialog shows only the current alias with no indication that there are other names. Successful completion of the delete process removes one alias <s>but does not inform the user that the item persists with the other names</s>. Same issue if user gets to +delete by clicking delete icon on Item History view.

Issue #280, opened 2012-11-03
Nicer editing of Extra MetaData (JSON)


ThomasWaldmann:
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/280) by Tarashish Mishra (Bitbucket: [sunu](https://bitbucket.org/sunu), GitHub: [sunu](https://github.com/sunu)).**

----------------------------------------

Currently the *Extra MetaData (JSON)* field showed while creating or modifying an item is just a text area with a bunch of json.

It should be revamped to make metadata editing easier.

RogerHaase:
closing, see #1067

PR #1066, opened 2020-08-27
<<rebuild index required>> improve whoosh indexes, fix #982

### LINKED ISSUE #982: Rework Whoosh indexes
Some ideas, needs more testing. Adding fields appears easier than changing or removing.

Whoosh indexes are defined in /storage/middleware/indexing.py line ~292. Changing will require users to rebuild indexes (./m index).

Additions to common fields:
* HAS_TAGS: BOOLEAN(stored=False),
  * true if item has any defined tags
  * response time of global tags will be improved because currently the whole index must be processed
* NAME_OLD: ID(stored=True),
  * history views can show old name for deleted items

Changed fields:
* SIZE: STORED(),
  * there are no searches based on size, so just store the value, no need to index
* TAGS: KEYWORD(stored=True),
* REPLY_TO: STORED(),
* CONTENT: TEXT(stored=False, spelling=True), -- needs more analysis
* ITEMLINKS: TEXT(stored=True),
* ITEMTRANSCLUSIONS: TEXT(stored=True),

Removed fields, these are not used:

        # Define dynamic fields
        dynamic_fields = [("*_id", ID(stored=True)),
                          ("*_text", TEXT(stored=True)),
                          ("*_keyword", KEYWORD(stored=True)),
                          ("*_numeric", NUMERIC(stored=True)),
                          ("*_datetime", DATETIME(stored=True)),
                          ("*_boolean", BOOLEAN(stored=True)),
                         ]

RogerHaase:
new: HAS_TAG, reduces tags view response time by ~75%
new: SUMMARYNGRAM, provide ngram searches of summary metadata
new: NAME_OLD, history and mychanges can show old name for deleted items
change: TAGS, change field type from ID to KEYWORD
change: SUMMARY, added field_boost=2.0


src/moin/storage/middleware/indexing.py
```diff
@@ -328,7 +334,10 @@ def __init__(self, index_storage, backend, wiki_name=None, acl_rights_contents=[
             # tokenized CONTENTTYPE from metadata
             CONTENTTYPE: TEXT(stored=True, multitoken_query="and", analyzer=MimeTokenizer()),
             # unmodified list of TAGS from metadata
-            TAGS: ID(stored=True),
+            TAGS: KEYWORD(stored=True, commas=True, scorable=True, field_boost=2.0),
+            # search on HAS_TAGS improves response time of global tags
+            # https://whoosh.readthedocs.io/en/latest/api/query.html?highlight=#whoosh.query.Every
+            HAS_TAG: BOOLEAN(stored=False),
```
ThomasWaldmann, line 340: ok, I've read that. ok, so HAS_TAG seems to be best practice for whoosh.

still not pretty imho having to do that, but assuming the docs are reflecting the current state of best practice, we can do it.

src/moin/items/__init__.py
```diff
@@ -693,6 +693,10 @@ def _save(self, meta, data=None, name=None, action=ACTION_SAVE, contenttype_gues
             meta[PARENTID] = currentrev.meta[REVID] if currentrev else meta[REVID]
             del meta[REVID]
 
+        if meta.get(TAGS, None):
+            # global tags uses this to search for items with tags
+            meta[HAS_TAG] = True
```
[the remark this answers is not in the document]
ThomasWaldmann: More ideas:

Does this make HAS_TAGS go into our item/revision storage?

How about only adding it right before indexing, so it won't be in our item storage, but just the indexer would see it?

Also: what about False values - don't we need them also?

```
meta[HAS_TAG] = bool(if meta.get(TAGS, None))
```

ThomasWaldmann (commented):
didn't do any whoosh or moin2 coding for a long time, so take it with a grain of salt.

src/moin/items/__init__.py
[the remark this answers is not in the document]
RogerHaase: The response time for tags view went from 4 seconds to 1 second on my 1.9.x wiki converted to 2.0 for testing. There is only 1 tag.
unique default items: 1100
top level default items: 160
default revisions: 8300
user profiles: 50
user revisions: 1300
wiki/data/default/data size: 155mb

src/moin/storage/middleware/indexing.py
```diff
@@ -354,21 +363,22 @@ def __init__(self, index_storage, backend, wiki_name=None, acl_rights_contents=[
...
 
         latest_revs_fields = {
             # ITEMID from metadata - as there is only latest rev of same item here, it is unique
             ITEMID: ID(unique=True, stored=True),
             # unmodified list of ITEMLINKS from metadata
-            ITEMLINKS: ID(stored=True),
+            ITEMLINKS: KEYWORD(stored=True, commas=True, scorable=True),
             # unmodified list of ITEMTRANSCLUSIONS from metadata
-            ITEMTRANSCLUSIONS: ID(stored=True),
+            ITEMTRANSCLUSIONS: KEYWORD(stored=True, commas=True, scorable=True),
```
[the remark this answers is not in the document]
RogerHaase: My only reason for changing is that after reading https://whoosh.readthedocs.io/en/latest/schema.html, KEYWORD seemed like a better fit than ID.  I could not find any difference in search behavior after the change. I will change both back to ID.

Not sure what to do about names that have embedded commas #580. Is it expected that names should have special parsing so commas are permissable inside quoted strings?

src/moin/storage/middleware/indexing.py
```diff
@@ -410,20 +420,6 @@ def __init__(self, index_storage, backend, wiki_name=None, acl_rights_contents=[
...
-                          ("*_numeric", NUMERIC(stored=True)),
-                          ("*_datetime", DATETIME(stored=True)),
-                          ("*_boolean", BOOLEAN(stored=True)),
-                         ]
-
-        # Adding dynamic fields to schemas
-        for glob, field_type in dynamic_fields:
-            latest_revisions_schema.add(glob, field_type, glob=True)
-            all_revisions_schema.add(glob, field_type, glob=True)
-
```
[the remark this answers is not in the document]
RogerHaase: OK, will change this back as well, but I have no idea how a user could add their own metadata and get it back in a useful way.

If you are thinking of the Extra Metadata (JSON) display on the item modify view, that was removed 6 months ago: #700. There is also #280 that should be closed (with no additional code changes).

src/moin/items/__init__.py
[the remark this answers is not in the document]
ThomasWaldmann: ok, that's substantial.

src/moin/storage/middleware/indexing.py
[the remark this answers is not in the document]
ThomasWaldmann: Well, when using ID, there is at least no indexing problem for item names with commas.

When thinking about what users enter for item names, I guess we have to expect anything that is valid unicode.

src/moin/storage/middleware/indexing.py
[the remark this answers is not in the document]
ThomasWaldmann: Well, the original idea was that users can enter own metadata via extra metadata and that this is indexed and can be quickly searched for. If it helps we can initially leave that away, but in general i think it can be a useful feature.

ThomasWaldmann:
did you check whether it is possible to have HAS_TAG just in the index, but not in moin storage metadata?

would be cleaner as this is a rather whoosh specific internal optimization.

RogerHaase:
I thought HAS_TAG was just in the index, but that is not true. Will fix.

For some reason, I cannot reply to this message on github, must use email.

src/moin/templates/ajaxsearch.html
```diff
@@ -56,14 +56,25 @@ <h4>{{ _("Please check these tickets if they cover your issue:") }}</h4>
...
             {% if results %}
                 <div class="moin-search-results">
-                    <table>
+                    {% set count = [0] %}
+                    <table
+                        {% set vars = {'old_acl_in': [None, None, None], 'old_permission': False} %}
                         {% for result in results %}
                             {% set fullname = result.namespace + '/' + result.name[0] if result.namespace else result.name[0] %}
-                            {% if user.may.read(fullname) %}
+                            {% set new_acl = result['acl'] if 'acl' in result else '' %}
```
ThomasWaldmann, line 66: could this be `result.get('acl', '')` ?

src/moin/storage/middleware/indexing.py
```diff
@@ -148,6 +148,23 @@ def get_names(meta):
...
+    :param names: item NAME from whoosh index, where NAME is a list
+    :return: parent names list
+    """
+    parent_names = []
+    for name in names:
+        parentname_tail = name.rsplit('/', 1)
+        if len(parentname_tail) == 2:
+            parent_name = parentname_tail[0]
+            if parent_name not in parent_names:
+                parent_names.append(parent_name)
```
ThomasWaldmann, line 164: is there some requirement for specific result order or for result being a list?

if not, you could also use a set, that would dedupe automatically.

src/moin/storage/middleware/protecting.py
```diff
@@ -181,6 +181,15 @@ def search_meta(self, q, idx_name=LATEST_REVS, **kw):
             if result:
                 yield rev
 
+    def may_read_rev(self, windex):
```
ThomasWaldmann, line 184: `windex` - can we use a better name than that?

it does not seem to be a whoosh index, but rather a whoosh query result.

src/moin/templates/ajaxsearch.html
```diff
@@ -56,14 +56,25 @@ <h4>{{ _("Please check these tickets if they cover your issue:") }}</h4>
...
+                            {% set new_parentnames = theme_supp.parentnames(result.name) %}
+                            {% set new_acl_in = [new_acl, result['namespace'], new_parentnames] %}
+                            {% if vars['old_acl_in'] == new_acl_in %}
+                                {# save a few seconds by avoiding recalculation of acls #}
+                                {% set permission = vars['old_permission'] %}
+                            {% else %}
+                                {% set permission = flaskg.storage.may_read_rev(result) %}
+                            {% endif %}
```
ThomasWaldmann, line 74: this looks a bit like a 1-entry LRU cache?

maybe rather use a 100 entry real LRU cache? iirc there are ready-to-use implementations of this in either moin2 or stdlib.

src/moin/themes/__init__.py
```diff
@@ -35,6 +35,7 @@
 from moin.utils.mime import Type
 from moin.utils import show_time
 
+
```
ThomasWaldmann, line 38: accidental?

src/moin/templates/ajaxsearch.html
RogerHaase: yes, fixed in my repo as the above caused traceback

src/moin/storage/middleware/indexing.py
RogerHaase: That code was just moved as was, will change to use set.

src/moin/storage/middleware/protecting.py
RogerHaase: It is a whoosh item index. The query results are passed to ajaxsearch.html as `results`, here ajaxsearch.html is in a loop processing the results and calling with a whoosh index.

src/moin/templates/ajaxsearch.html
RogerHaase: Opened #1079. The ACL code seems difficult to grok.

src/moin/themes/__init__.py
RogerHaase: yes, will fix

Issue #1069, opened 2020-09-05
ACL validation is slow


RogerHaase:
Review fixes made to Global Index, History, Tags are sluggish #946, Item History is sluggish #940, and Rendering of search results is slow #1068. 

A theory is ACL processing was written with item read/save as the primary focus. Reports  that can be run using only the selected list of Whoosh index data need an alternative process that does not read wiki content nor wiki metadata. 

There is some code that refers to Whoosh index data as metadata or meta. The index data and wiki metadata are similar, but not the same. Naming the index data more clearly will avoid future confusion.

In /storage/protecting.py there is this code to workaround slow ACL processing. With the workaround enabled, it takes 62 seconds to process 1002 Global History items (as a single page). With the workaround disabled, it takes 137 seconds.:

    def search_meta(self, q, idx_name=LATEST_REVS, **kw):
        """
        Yield an item's metadata, skipping any items where read permission is denied.

        The intended use of this method is to return the current rev metadata for all
        of the items in namespace subject to query restrictions. This is useful for reports
        such as Global Index, Global Tags, Wanted Items, Orphaned Items, etc.

        Save processing time by avoiding a full ACL check when the answer will be the same as the last.
        """
        last_rev_acl_parts = (None, None, None)
        last_rev_result = None
        for rev in self.indexer.search_meta(q, idx_name, **kw):
            rev = ProtectedItemMeta(self, rev)
            this_rev_acl_parts = (rev.meta[NAMESPACE], rev.meta.get(PARENTNAMES), rev.meta.get(ACL))
            if this_rev_acl_parts == last_rev_acl_parts:
                # skip the acl check because we know the answer will be the same
                if last_rev_result:
                    yield rev
                continue
            last_rev_acl_parts = this_rev_acl_parts
            result = rev.allows(READ) or rev.allows(PUBREAD)
            last_rev_result = result
            if result:
                yield rev

There is similar code in the search_meta_page method of protecting.py. The time to process a page of 50 MyChanges is 2.7 seconds with the code enabled, 5.2 seconds with the code disabled.

In /templates/ajaxsearch.html there is similar code that saves 3 seconds of processing time when rendering 100 search hits.:

                    <table
                        {% set vars = {'old_acl_in': [None, None, None], 'old_permission': False} %}
                        {% for result in results %}
                            {% set fullname = result.namespace + '/' + result.name[0] if result.namespace else result.name[0] %}
                            {% set new_acl = result.get('acl', '') %}
[...]

PR #1074, opened 2020-09-24
add lru cache to protecting.py ProtectingMiddleware _allows method #1069

RogerHaase:
skip ACL permission recalculation when an item with same
user_names, acls, parentnames, namespace, and right is in lru cache

improves response times for reports that process all/many item revisions

removed ProtectedItemMeta class from protecting.py, moved code to
ProtectingMiddleware class

broke one test that was testing dead code, removed test from test_item.py,
themes/__init__.py subitem_index, and items/__init__.py get_mixed_index


src/moin/storage/middleware/protecting.py
```diff
@@ -73,6 +85,10 @@ def __init__(self, indexer, user, acl_mapping):
         self.eval_acl = lru_cache_decorator(self._eval_acl)
         lru_cache_decorator = lru_cache(LOOKUP_CACHE)
         self.get_acls = lru_cache_decorator(self._get_acls)
+        lru_cache_decorator = lru_cache(ACL_CACHE)
+        self.allows = lru_cache_decorator(self._allows)
+        # placeholder to show we are passing meta data around without effecting lru caches
```
ThomasWaldmann, line 90: typo "affecting"?

ThomasWaldmann: also not sure i understand that...

it seems you have remove usage of rev.meta and rather use rev as a dict-like object.

guess that is somehow departing from the initial idea of middlewares somehow exposing the same or at least similar interface, so that one can use the middlewares or not.

but, besides that, if the idea is not to use .meta any more, shouldn't it rather be gone and raise an AttributeError?

src/moin/storage/middleware/protecting.py
```diff
@@ -162,36 +177,65 @@ def search_meta(self, q, idx_name=LATEST_REVS, **kw):
         The intended use of this method is to return the current rev metadata for all
         of the items in namespace subject to query restrictions. This is useful for reports
         such as Global Index, Global Tags, Wanted Items, Orphaned Items, etc.
-
-        Save processing time by avoiding a full ACL check when the answer will be the same as the last.
         """
+        from . indexing import parent_names
```
ThomasWaldmann, line 181: does that work? usually one writes `from .indexing ...`.

src/moin/storage/middleware/protecting.py
```diff
@@ -162,36 +177,65 @@ def search_meta(self, q, idx_name=LATEST_REVS, **kw):
...
-    def may_read_rev(self, windex):
+    def may_read_rev(self, meta):
         """
         Return true if user may read item revision represented by whoosh index hit.
-        Called by search templates, maybe others.
+        Called by ajaxsearch template.
         """
-        hit = ProtectedItemMeta(self, windex)
-        may_read = hit.allows(READ) or hit.allows(PUBREAD)
+        from . indexing import parent_names
```
ThomasWaldmann, line 198: same here

ThomasWaldmann: also this method seems to have duplicate code as already used above, so maybe above code could just call this method?

src/moin/storage/middleware/protecting.py
```diff
@@ -162,36 +177,65 @@ def search_meta(self, q, idx_name=LATEST_REVS, **kw):
...
+        :param user_names: user names to use for permissions check (default is to
+                          use the user names doing the current request)
+        :rtype: bool
+        :returns: True if you have permission or False
+        """
+        acl_cfg = self._get_configured_acls(self.fqnames[0])
+        for user_name in user_names:
+            for full_acl in self.full_acls():
+                allowed = self.eval_acl(full_acl, acl_cfg['default'], user_name, right)
+                if allowed is True and pchecker(right, allowed, self.meta):
```
ThomasWaldmann, line 233: if allowed and ...

src/moin/storage/middleware/protecting.py
```diff
@@ -162,36 +177,65 @@ def search_meta(self, q, idx_name=LATEST_REVS, **kw):
...
+                    return True
+        return False
+
+    def search_meta_page(self, q, idx_name=LATEST_REVS, pagenum=None, pagelen=None, **kw):
+        """ mychanges response time
```
ThomasWaldmann, line 238: ??

src/moin/storage/middleware/protecting.py
```diff
@@ -203,21 +247,15 @@ def search_meta_page(self, q, idx_name=LATEST_REVS, pagenum=None, pagelen=None,
...
-            if this_rev_acl_parts == last_rev_acl_parts:
-                # skip the acl check because we know the answer will be the same
-                if last_rev_result:
-                    yield rev
-                continue
-            last_rev_acl_parts = this_rev_acl_parts
-            result = rev.allows(READ) or rev.allows(PUBREAD)
-            last_rev_result = result
+        # import here to avoid circular import error
+        from . indexing import parent_names
```
ThomasWaldmann, line 251: see above, please do a global change to fix all of these.

ThomasWaldmann (commented):
this was bad to review due to the amount of changes all done in 1 commit.

also i guess the protectingmiddleware is not optional any more due to the interface change?
(not sure if it was before though)

src/moin/storage/middleware/protecting.py
RogerHaase: I have given up on using windex instead of meta, seems meta is used everywhere as the name for whoosh index data.

Will fix the use of rev/revs vs. meta/metas in frontend/views.py and admin/views.py in next commit. So revs will be used for Items, metas for whoosh indexes (dict).

Several reports (History, Index, Tags, MyChanges...) require only whoosh index data. These run much slower (10-20 times slower, see #946) if Item instances (with an open data file) are returned. If that is exposing a dissimilar interface, it is necessary.

Will fix affecting.

src/moin/storage/middleware/protecting.py
RogerHaase: will fix

src/moin/storage/middleware/protecting.py
RogerHaase: will fix

src/moin/storage/middleware/protecting.py
RogerHaase: will fix

src/moin/storage/middleware/protecting.py
RogerHaase: oops, will fix

PR #1139, opened 2021-11-03
html validation error, malformed uri reference; fixes #1137

### LINKED ISSUE #1137: html validation error, malformed uri reference
Links with non Ascii characters are not properly escaped. Add the following to a moinwiki item and run html validator. 

[[AjudaNaFormatação]]

RogerHaase:


RogerHaase:
breaks [[#mylink]]

Issue #1082, opened 2020-12-08
move CI to github actions


ThomasWaldmann:
about: https://news.ycombinator.com/item?id=25338983

example: https://github.com/borgbackup/borg/blob/master/.github/workflows/ci.yml

RogerHaase:
CI build not running after last pull request.

Went to travis-ci.com, signed up, ordered a free account, received 10,000 credits for MoinMoin Wiki Engine.

There is no option to rerun build on pull request page:  https://github.com/moinwiki/moin/pull/1106

Made a few CSS changes, amended last commit and pushed. CI build did not restart as before. No obvious way to restart CI build.

Not sure what I did, now I get a warning message:

    Builds have been temporarily disabled for private and public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance.

    -190 available credits (purchase date: March 18, 2021)
    You have used 10190 of 10000 credits
    -190 available

ThomasWaldmann:
10000 sounds much, but isn't. I just moved away from travis-ci for other projects.

RogerHaase:
Assuming the start to fixing this issue is to remove all Travis CI stuff...

Following https://remarkablemark.org/blog/2021/03/15/github-remove-travis-ci/#:~:text=Revoke%20Travis%20CI%20from%20Account%20settings%20%3E%20Applications%3A,settings%2C%20check%20if%20Travis%20CI%20is%20added%20in%3A

Under Organization, Travis CI must be removed from 3rd party apps and github apps.  Nothing to do under Webhooks.

Under Repository Config, delete .travis.yml. Nothing to do under Branches and Webhooks.

PR #1185, opened 2022-02-14
remove Travis CI .travis.yml file; #1082

RogerHaase:


.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
ThomasWaldmann, line 21: btw, for borgbackup 1.2 (to be released soon) i decided to drop everything < 3.8.

even the stable dists of misc OSes now have 3.8+ and usually software that is not released yet (== released in the future) does not get into current-stable, but rather into next-stable OS release.

also means:
- less load on / less waiting for CI server.
- ability to use recent python features

.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest
+        pip install -r requirements.d/development.txt
+        pip install .
```
ThomasWaldmann, line 34: not sure, but i could imagine that `pip install -e .` is faster (because code files stay in git workdir and are not copied).

.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,43 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+
+    steps:
+    - uses: actions/checkout@v2
```
ThomasWaldmann, line 24: https://github.com/borgbackup/borg/blob/master/.github/workflows/ci.yml#L78

have a look there ^^^

you can also take some other stuff from there, e.g. running the linter only once and before the real testing begins.

.github/workflows/github-actions.yml
RogerHaase: fixed in try 2

ThomasWaldmann:
```
            - os: macos-10.15  # macos-latest is macos 11.6.2 and hanging at test_fuse, #6099
```

I guess you don't need FUSE for moin, so you could also use macos-latest.

.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,71 @@
+# Modified for MoinMoin, copied from: https://github.com/actions/starter-workflows/blob/main/ci/python-package.yml
+# and https://github.com/borgbackup/borg/blob/master/.github/workflows/ci.yml
```
ThomasWaldmann, line 2: maybe also rename this file to ci.yml, github-actions.yml is too generic.

also, non-ci stuff would be better put into a separate file.

ThomasWaldmann:
I guess after merging this PR, you'll see the github CI run for the first time and likely discover the need for more changes / fixes. So it might make sense merging this soon.

PR #1192, opened 2022-02-18
require python 3.8+; fixes #888

### LINKED ISSUE #888: require python >= 3.6
we should raise requirements asap:

- after people (including tw) moved from debian stretch (py35) to buster
- to get rid of too many versions which slow down travis-ci
- to upgrade travis to bionic, see #859 
- to use new py36 features https://docs.python.org/3/whatsnew/3.6.html

RogerHaase:


UlrichB22:
In addition to the changed objects I found further references to python 3.5:
moin.spec
quickinstall.py
docs/admin/install.rst
docs/admin/requirements.rst
src/moin/__init__.py
Would you like me to support you?

setup.py
```diff
@@ -9,8 +9,8 @@
 from setuptools import setup, find_packages
 
 
-if sys.hexversion < 0x3050000:
-    sys.exit("Error: MoinMoin requires Python 3.5+., current version is %s\n" % (platform.python_version(), ))
+if sys.hexversion < 0x3080000:
+    sys.exit("Error: MoinMoin requires Python 3.8+., current version is %s\n" % (platform.python_version(), ))
```
ThomasWaldmann, line 13: this is quite old code (not possible otherwise many years ago). but since a while, there is a better method:

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

RogerHaase:
A PR to fix those 3.5 references would be helpful. Thanks for offering.

PR #1198, opened 2022-02-26
Fix PageList migration to use only regexp arguments

abirke:
The moin-1.9 PageList macro passed all arguments to the FullSearch macro which basically means all arguments were handled as regular expressions. This Pull Request fixes the PageList macro migration to only use regular expression arguments for the converted ItemList macro. It also fixes a typo in the converted macro and adds some basic testing for the migration procedure.

### DIFF
--- a/src/moin/scripts/migration/moin19/macros/PageList.py
+++ b/src/moin/scripts/migration/moin19/macros/PageList.py
@@ -17,14 +17,19 @@ def convert_page_list_macro_to_item_list(node):
     The moin1.0 PageList macro used to pass all arguments to the FullSearch
     macro, so they were essentially all treated as regular expression search queries.
     After conversion to the ItemList macro, the argument will be a simple "regex" argument.
+    To make sure all available pages are searched (not only subpages of the current page)
+    the item parameter is set to empty string (item="").
 
     Example conversions:
 
-    | PageList macro (moin1.9)       | ItemList macro (moin2)           |
-    |--------------------------------|----------------------------------|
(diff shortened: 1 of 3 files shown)


src/moin/scripts/migration/moin19/macros/PageList.py
```diff
@@ -12,15 +14,16 @@
...
+    After conversion to the ItemList macro, the argument will be a simple "regex" argument.
 
     Example conversions:
 
     | PageList macro (moin1.9)       | ItemList macro (moin2)           |
     |--------------------------------|----------------------------------|
     | <<PageList()>>                 | <<ItemList()>>                   |
-    | <<PageList(SomeSubPage)>>      | <<ItemList(item="SomeSubPage")>> |
+    | <<PageList(SomeSubPage)>>      | <<ItemList(regex="SomeSubPage")>> |
     | <<PageList(regex:Rnd[^abc]+)>> | <<ItemList(regex="Rnd[^abc]+")>> |
```
RogerHaase, line 27: ItemList macro displays matched subpages by default, PageList displays all matched pages. To make ItemList work like PageList, I think ItemList needs `item=""` parameter.

| <<PageList(Help)>> | <<ItemList(item="",regex="Help")>> |

abirke: Good point, thanks for pointing it out. I changed the migration accordingly and hope the converted macros now behave the same as in version 1.9.

RogerHaase:
Thanks for helping

Issue #1222, opened 2022-04-05
ItemList macro does not support namespaces


RogerHaase:
See MoinWikiMacros help page.

Try this macro on an item not in default namespace : `<<ItemList(item="help-en/OtherTextItems", startswith="P")>>`

Items are found by search, but links fail because they lack namespace.

UlrichB22:
I would appreciate it if the macro logic could be switched to the search engine because we have an impact on performance in the actual implementation. The same applies to the MonthCalendar macro. We are running a wiki for more than 10 years with a calendar. The calendar item has thousands of children. 
Maybe I can help if you give me some hints and suggestions.

RogerHaase:
Please point to the line numbers in ItemList.py and MonthCalendar.py that you think are causing the slowdown.

UlrichB22:
In ItemList starting from 
https://github.com/moinwiki/moin/blob/5221b0453f90386a726bcb7b0888861d20fa73b6/src/moin/macros/ItemList.py#L131
up to line 140.

In MonthCalendar this runs in a loop for every day of month:
https://github.com/moinwiki/moin/blob/5221b0453f90386a726bcb7b0888861d20fa73b6/src/moin/macros/MonthCalendar.py#L261
I would prefer to run one whoosh range query with `[<parentitem>/YY-MM-01 TO <parentitem>/YY-MM-31] ` before the loop and just check if each day is in the result list inside the loop.

RogerHaase:
OK, will look at this more in a couple of days. Currently stuck on another problem.

In previous work at speeding up code, I found that Whoosh calls that return a generator take 0 time.  Retrieving an revision from the generator takes time. Much time is saved if only the Whoosh index data is returned, without the content attribute that contains an opened file to read the item content.

Most of the code that calls a variable "meta" is actually the whoosh index. One exception is when an item is updated there is a new "meta" file written to storage. The Whoosh index data (aka "meta") includes the item content - I think the data is mangled for things like PDF or image files so Whoosh can index the content, not sure how mangled text files are. See storage/middleware/indexing.py near line 396 - the comment above it says converted to text/plain. 

I have wondered what would happen if the content data were not stored. Seems like a waste to store content in the index and not use it.

See closed issue Global Index, History, Tags are sluggish #946

UlrichB22:
Following changes seem to fix the primary issue:

Change pagename to pagename.fullname here and in line 107 
[...]

PR #1230, opened 2022-04-20
MonthCalendar: optimize performance using index search

UlrichB22:
This is just an example how to optimize macro runtime mentioned in #1222. 
Requires more discussion before merging. The search_names() function should be moved to another module for reusability.
Reduces runtime from 14 seconds to 3 seconds in a real example for a one year calendar.

### DIFF
--- a/src/moin/macros/MonthCalendar.py
+++ b/src/moin/macros/MonthCalendar.py
@@ -222,7 +222,7 @@ class Macro(MacroInlineBase):
 
         # get list of calendar items for given month
         item_month = "{:s}/{:4d}-{:02d}-".format(parmpagename[0], year, month)
-        date_results = search_names(item_month)
+        date_results = search_names(item_month, limit=100)
 
         r7 = range(7)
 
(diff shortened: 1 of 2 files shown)


RogerHaase:
Take a look at moving the search_names procedure to storage/middleware/indexing.py. Probably near the top with the other procedures.

RogerHaase:
I tried a few things but could not find a means of speeding this up. The whoosh Prefix function seems very fast. I made a note to myself to try Prefix within +index.

Maybe use a list comprehension instead of the for loop in the search_names procedure

    result_names = [x[NAME] for x in results]

or to add functionality and maybe a bit slower, a dict comprehension

    results = {x[NAME][0]: x[CONTENT][:50] for x in results}

that way, with a bit more code a mouseover tooltip could be added to show the 1st 50 characters of the day's content. And we finally have a use for having content in the Whoosh index.

UlrichB22:
Thanks for the suggestions. I will keep the idea about the tooltips in mind for a later extension.

PR #1238, opened 2022-05-06
Allow to run "./quickinstall.py" whithout any other arguments

metemaddar:
Whenever we ran `./quickinstall.py` without any arguments it should be the same as `python quickinstall.py`. But we got the following error:

```
  File "./quickinstall.py", line 746, in <module>
    choice = 'cmd_%s' % args[1]
IndexError: list index out of range
```

Added if statement to check if length of args equals to 1, then runs the default command.

### DIFF
--- a/quickinstall.py
+++ b/quickinstall.py
@@ -696,6 +696,11 @@ if __name__ == '__main__':
             choices.add(name)
     args = sys.argv[:]
 
+    if len(args) == 1:
+        # User run command as `./quickinstall` whithout any other arguments
+        command = getattr(commands, 'cmd_quickinstall')
+        command()
+
     if len(args) > 2 and args[-1] == '--help':
         # m and m.bat have trailing --help so "./m" comes across as "python quickinstall.py --help"
         # if user did "./m <option>" we have "python quickinstall.py <option> --help" then we can delete the --help and do <option>
--- a/quickinstall.py
+++ b/quickinstall.py
@@ -747,7 +747,6 @@ if __name__ == '__main__':
 
             else:
                 # we have some simple command like "./m css" that does not update virtualenv
-                print('lenght of args: ',len(args))
                 choice = 'cmd_%s' % args[1]
                 choice = choice.replace('-', '_')
                 if choice in choices:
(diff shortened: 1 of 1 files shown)


RogerHaase:
I could not reproduce the error, but found other errors caused by new releases of packages.

Please run "./m quickinstall" again and open an issue if you still have problems. Also add a bit more information about your OS, and how long you have been running moin2.

metemaddar:
Os: Ubuntu
To reproduce the error, run:

`./quickinstall`

It was the first time I ran the moin2 and would like to participate in development faced this error. I was stucking at the installation stage, because I didn't expect to get this error.

This error acures because we have no checking the condition where we have just `arg[0]` and no `arg[1]`

metemaddar:
The problem didn't resolve in first commit. So moved the if statement as `elif` to previous command decision statement.

RogerHaase:
The correct command sequence is to activate the environment, then

    ./m quickinstall

where the "m" executes a script that passes an argument. To see what else "m" can do:

    ./m

RogerHaase:
https://moin-20.readthedocs.io/en/latest/admin/install.html#installation-for-developers

metemaddar:
Thank you very much for clarification and the reference. If I get well, we have decided to replace the `quickinstall.py` command with `m` which will a more clear and relevant name to the moinmoin.
Of course just to make my opinion more clear I would like to say that as quickinstall.py has shabang in the begining and it is a executable file, it can be run as standalone. So it would be good to decide what happens when we run it without trailing arguments.
Maybe it is better to show help text or do the default behaviour.
I think as we have mentioned in the [document which you provided as well](https://moin-20.readthedocs.io/en/latest/admin/install.html#installing) , to run the comman `python quickinstall.py ` to create v-env and install packages, it would be a right decision to let quickinstall.py to do this behaviour as well.

metemaddar:
> The correct command sequence is to activate the environment, then
> 
> ```
> ./m quickinstall
> ```
> 
> where the "m" executes a script that passes an argument. To see what else "m" can do:
> 
> ```
> ./m
> ```

Although it is completely right, but we have a more step before to create the v-env and we should run as its mentioned in [#installing section of the document](https://moin-20.readthedocs.io/en/latest/admin/install.html#installing) :

`python3 quickinstall.py `

Issue #1258, opened 2022-06-01
Wrong itemlinks if using relative path in namespaces


UlrichB22:
Relative link paths are not converted correctly in meta data.

E.g. the item help-en/Home uses following link

` * [[../moin|Moin Wiki markup item]]`
 
In metadata this is shown as moin instead of help-en/moin. 
As a result, the help-en/moin element is included in the list of orphans:

http://localhost:8080/+orphans

The Code is around this block:

https://github.com/moinwiki/moin/blob/136ee26cb16f00d6965cce278e80d1a39f2b9eaa/src/moin/storage/middleware/indexing.py#L324

Need some help to fix.

RogerHaase:
The metadata for help-en/moin has a name "moin" and a Namespace "help-en" so it is correct.

I think the problem is in apps/frontend/views.py ~1740. "def orphaned_items" and its helpers does not support namespaces.

UlrichB22:
To reproduce I used a new empty wiki with only help-en and help-common loaded. If you look at meta data http://localhost:8080/+meta/help-en/Home you will see a lot of invalid links in the list of Item Links.

According to http://localhost:8080/help-en/moin every link with a simple itemname without slashes is located in default namespace. Every item link for the same namespace is preceded with the namespace name. This implies, that relative paths to siblings need to be translated to namespace/itemname within metadata.  

The list of orphans uses the itemlink meta data from index only, this is perfect.

RogerHaase:
I misunderstood, agree the metadata for Item Links and Item Transclusions is incorrect - each link should have the complete path.

RogerHaase:
I think a possible fix for Item Links is in /storage/middleware/indexing.py near line 1159.  But still testing. Also finding doc errors in moin help page.

    def preprocess(self, meta, data):
        """
        preprocess a revision before it gets stored and put into index.
        """
        # ~ content = convert_to_indexable(meta, data, self.name, is_new=True)
        content = convert_to_indexable(meta, data, self.fqname.fullname, is_new=True)
        return meta, data, content

RogerHaase:
The above fix creates another problem. After loading the help-en namespace in a new wiki, the show meta for +meta/help-en/moin displays Item Links as:

[...]

PR #1260, opened 2022-06-08
Consider namespace in convert_to_indexable

UlrichB22:
Possible Solution for #1258. 
Needs further testing.


RogerHaase (commented):
This seems to work. The only issue I see is on line 266: meta[NAMESPACE] is not None

I cannot recall ever seeing a meta[NAMESPACE] of None. It is always an empty string: ""

UlrichB22:
will be replaced by new PR.

PR #1303, opened 2022-08-08
Fix quickinstall not working if called with path

JanHoefelmeyer:
Instead of checking for exactly 'quickinstall.py', I'm now checking if exactly 1 argument was called that ends with quickinstall.py. This should include all relative paths. 

Note that being in the main folder seems to still be necessary, as e.g. calling ..\\..\quickinstall.py won't throw errors but won't work either, however this is not a result from this change (but was already this way before) and is likely much less relevant if users follow the quickinstall instructions.

### DIFF
--- a/quickinstall.py
+++ b/quickinstall.py
@@ -740,7 +740,7 @@ if __name__ == '__main__':
             command()
 
         else:
-            if args == ['quickinstall.py']:
+            if (len(args) == 1 and  args[0].endswith('quickinstall.py')):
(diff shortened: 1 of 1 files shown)


RogerHaase:
It is easy to miss this error in the test output. The other possibility is GitHub has a later version of flake8. Check

    moin\.tox\flake8\Lib\site-packages\flake8...dist-info

mine is at 5.0.2. Flake8 version 5.0.0 started flagging no space after keyword as errors.

Please push a correction.

Thanks for helping!

Issue #1311, opened 2022-08-25
Quickinstall not working on Ubuntu 22.04 Lts Desktop


JanHoefelmeyer:
I've tried setting up Moin2 on Windows 10 and Ubuntu 20.04 (lts) with no problems. Trying to set it up on Ubuntu 22.04 (lts) fails with no error messages given. The venv is supposedly created, but trying to use any further commands (e.g. sample) just leads to an error telling me that there is no such command.

Edit: I should mention that the entire log is empty. Even the messages indicating a successful creation are not present.

UlrichB22:
I tested with a fresh installed Ubuntu 22.04 Server version. 
quickinstall.py has complained about missing packages, you find the full log output in a file called m-quickinstall.txt.
After running
```
sudo apt-get install python3.10-venv
sudo apt-get install python3-pip
```
quickinstall was running successful. Afterwards you need to activate the venv and can proceed with instance creation and more as described in https://moin-20.readthedocs.io/en/latest/admin/install.html?highlight=venv#creating-a-wiki-instance .

UlrichB22:
I can reproduce with the desktop version. Can be further analysed using pycharm ->   https://moinmo.in/MoinMoin2/IdeSetup .

JanHoefelmeyer:
Weirdly enough, trying to execute the script with pycharm seems to work differently. I got the error message of ensurepip not being enabled. After running 
```sudo apt-get install python3-venv```
trying to run quickinstall.py still did not seem to work, as it was once again completed without logs in 0:00:00.

However, running it through the debugger apparently did work, as it took 30 seconds and finished with the correct messages. After that, I am now able to create and run a wiki.

UlrichB22:
Problem seems to be related to recursively opening the logfile and redirecting stdout/stderr a 2nd time. Can you check this fix:
https://github.com/UlrichB22/moin/commit/917a3e332e12ee60c26861a91983f573f0a5362f

Works for me on Ubuntu 22.04 desktop. Need to be also tested on Windows, I have no such environment.

JanHoefelmeyer:
Works on Windows. Need to test it on Ubuntu still, but I seem to have a problem with venv atm (unrelated to moin, since I can't create a venv even without the skript at the moment.) ((Error when creating venv, Error: Command '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1) despite venv and pip being installed. Will try to fix this error and then ammend or edit this comment.)

[...]

PR #1312, opened 2022-09-03
quickinstall.py: remove recursive log file open

UlrichB22:
When quickinstall.py is run with parameter --FirstCall the redirect of stdout and stderr is already done. No need for a second redirect to the same logfile. Obviously this is a problem with Ubuntu 22.04 LTS desktop version.

Tested with Ubuntu 22.04 desktop and Debian 11 Server version. Please assist with tests on Windows.


UlrichB22:
Fixes #1311

PR #1323, opened 2022-12-18
Feature/basic theme

billyquith:
Modified the basic theme. Looks a little easier on the eye IMHO. Could do with another couple of passes to make the whole looks consistent. 

image

### DIFF
--- a/src/moin/themes/basic/static/css/theme.css
+++ b/src/moin/themes/basic/static/css/theme.css
@@ -5882,6 +5882,17 @@ button.close {
 ul.moin-pull-right {
   float: right !important;
 }
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  padding-left: 0.5em;
+  background-color: #E9EAEC;
+  border-radius: 7px;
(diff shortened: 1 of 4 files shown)


RogerHaase:
Thanks for helping!

I like your changes. Only wish is the text on the sidebar (Site Navigation, etc) should have more contrast.

RogerHaase:
Topside is the default theme on moin2, Modernized is default on moin1.x.

Long ago the Modernized theme was using Stylus. I removed the Stylus code ~2020-08-06 after the implementation of CSS variables. 

I never was a Bootstrap fan, and wasted much time chasing Stylus bugs. My main complaint is the size of the resulting CSS files: 134k for the basic theme.css; 9k for modernized; 7k for topside. iirc the reason we are using Bootstrap 3.1.1 is because many of the variable names changed in 4.0.

My preference is to limit use of Bootstrap to the Basic theme and enhance the use of CSS Variables for the other themes. But if you have a strong argument for use of Bootstrap, let's hear it.

Issue #1259, opened 2022-06-03
/+meta/help-en/markdown does not show any Item Links


RogerHaase:
Same problem with `+meta/help-en/rst`

All markdown items have this problem, including the sample item.

UlrichB22:
In `/help/en/rst.meta` the itemlinks are included in the externallinks list:

```
"externallinks": [
    "http:Home",
    "http://www.python.org/",
    "http://moinmo.in/",
    "http:/users/Home",
    "http:rst/MissingItem",
```

UlrichB22:
Result in meta is now:

```
External Links: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#enumerated-lists, http://moinmo.in/, http://www.python.org/
Item ID: 412d0e59c79147058ef2e947d1772d5f
Item Links: [help-en/rst/Home](http://localhost:5000/help-en/rst/Home), [help-en/rst/users/Home](http://localhost:5000/help-en/rst/users/Home), [Home](http://localhost:5000/Home), [rst/MissingItem](http://localhost:5000/rst/MissingItem)
```

Links moved from External links to Item links, but target seems to be wrong. E.g. first link is 
http:/Home with the comment `link to a page in the default namespace of this wiki
`but it is pointing to http://localhost:5000/help-en/rst/Home

RogerHaase:
Agree the help pages need more help. Moving the help items from the default namespace to the help-en namespace made some of the examples confusing and wrong.

There is this clause near the top: `The examples below are dependent upon the use of the "help-en" namespace to separate help items from the main wiki content. Some target items may be missing from the default or users namespace.`

The sibling link should be

    |`[sibling item](../Sibling)`|[sibling item](../Sibling)|

markdown and moinwiki syntax both use a leading / to link to a subitem

Will open a new issue for fixing the help docs

bylsmad:
I did a bit of work on the help files already see #1373

RogerHaase:
TODO: As of today, meta views for docbook and html items still show "Item Links: (None)".

PR #1354, opened 2023-03-05
add missing itemlinks in rst item's metadata #1259

RogerHaase:
add tests for mailto, wiki.local


RogerHaase:
there is better way to fix missing size, sha1

PR #1355, opened 2023-03-06
167 wiki root

bylsmad:
wiki_root now added to data_href in converter.links

docbook_in now converting transclusions into xinclude:include instead of html.object

included site crawl test which is passing on current sample wiki with couple expected failures (TODO: fix the sample wiki)

included dump function which was useful in displaying xml while debugging


UlrichB22:
Thanks for supporting moin. Can you please split this into two independent pull requests for the converters module and for scrapy? This would help to test and merge the converters module first. I also have a change for docbook_in.py.

I have no experience with Scrapy. The tests failed due to module dependencies. This needs to be investigated further because it is only required by developers AFAIK.

bylsmad:
new pull request with the scrapy test #1362 which is now passing in the github action on macos and ubuntu, currently skipped on windows as it did not complete after 20 minutes

Issue #729, opened 2018-10-13
move from flask-script to flask.cli


ThomasWaldmann:
`flask-script` seems deprecated, `flask.cli` is built into flask, based on `click`.

http://flask.pocoo.org/docs/1.0/cli/

RogerHaase:
Closing as dup of #1109 

Jinja2, flask, and werkzeug must be upgraded together to remove flask-script

UlrichB22:
Maybe you can add a new label `flask 2.x` and keep this issue open as one of the subtasks.

RogerHaase:
I meant to close this, must have clicked the wrong button. I am working on #1109, but slowly.

In addition to Jinja2, flask, and werkzeug there are upgrades to CLI, markupsafe, itsdangerous, maybe others.

UlrichB22:
There is a prototype using flask cli in branch [github.com/UlrichB22/moin/tree/cli_preview](https://github.com/UlrichB22/moin/tree/cli_preview)

It offers the old flask-script based command `moin` and the new cli based command `moin2`.

This is only for testing and comparing the behavior of old and new version. For the final version I will rename the script module to cli and apply the changes. This will keep the history of each module. Command name will be `moin` again.

Migration to flask cli is not finished yet. Some commands and options like '--config' are still missing. I am still working on it. 
Please start with `moin2 --help` to find the available subcommands.

The main issue is fixing the codebase outside of the cli module. There is no request context and the flask cli seem to be more strikt in checking it. For example there is a function in converters/html_in that uses flash() to raise an error in the gui. Also app.py is mainly intended for the web application and needs some workarounds if called from cli.

Using a lot of try-blocks is not the preferred solution. I appreciate your comments and help. We can use the moin-devel mailing list for discussions. Please check, test and support.

RogerHaase:
Or we could make notes here. I have cloned your repo. The "html" item is missing from the sample data, but adding an invalid tag to create a flash message in "help-en/html" works for me. Ditto for the flash messages issued by the "markdown_in" converter.

Will test more tomorrow.

UlrichB22:
All commands inclusive import19 are now available, please see `moin2 --help`. The only problem is `dump-html`, it is broken because it relies too often on a valid session context. If you get a chance, can you help and try to fix it?

There are some minor changes or open topics:
- there is a new default flask command routes. I don't know if we should hide or deactivate and how to do it.
[...]

PR #1361, opened 2023-03-09
Move from flask-scripts to cli

UlrichB22:
This is the major step for #729


RogerHaase:
I found a couple of things that were broken before your changes, will open issues for these.

`moin storage-create --help` seems broken

`moin import19 -s -i -d ...`  does not initialize the index nor the storage; must do index-create first

Agree we can drop the `moin moin` command and just use `moin run`

still testing...

UlrichB22:
Thanks for testing. 

The options -i -s were mysterious before. They had no effect in import19 command. I removed it for import19. 
For a newbie it is hard to explain that on the command index-create you have to use the option -i to create the index. It is obsolete now. I will change the docs accordingly.

I think there is no command storage-create, it is just the long name of the option -s. To make it more complex the options have been stored in create_storage and create_index. This is not possible with flask cli. These variables have just been used as global variables before.

If it is ok I will cleanup the code related to options -i -s at a later time.

RogerHaase:
ok to clean up the -i -s mess later. I think this is pull request is good to go whenever you are ready.  Thanks for all the help.

PR #1363, opened 2023-03-11
CLI: update docs and comments

UlrichB22:
Updated documentation after migration from flask-script to cli, see #729.

Most changes have been made in `docs/man/moin.rst`.
I am aware that there are different opinions which cli interface is the better or newer one. Removed the words newer and older. For me the ./m interface is needed by developers and the moin cli is the only interface for normal wiki users. Hope this change is ok.


RogerHaase:
Thanks for all the work in getting cli, werkzeug, jinja, and flask up to date!

Issue #1387, opened 2023-04-03
create tests for cli


bylsmad:
working on pytests for cli in order to have confidence I'm not breaking anything while fixing #1378

UlrichB22:
Sounds great. You can raise a PR if you have the first test module available. Maybe I can support with some tests.
Please be aware that `moin dump-html` is still broken, that's one of the open topics from #729. Any help is appreciated.

bylsmad:
@UlrichB22 pull req #1389 is up, let's coordinate here by posting a comment when we start on a test for a command so we don't duplicate...I'm planning to turn back to #1378 for the time being as I believe the tests I need for it are in place

UlrichB22:
If you use MOINLOGGINGCONF to apply special logging options and turn debug on some tests will fail.
E.g. `pytest src/moin/cli/_tests/test_modify_item.py` reports

```
    def read_index_dump_latest_revs(out: str):
        """parse output of moin dump-index yielding the items in latest revs
    
        :param out: stdout of `moin index-dump --no-truncate` command
        :return: list of dicts with key value pairs from output"""
        [...]
            space_index = line.index(' ')
>           item[line[0:space_index]] = eval(line[space_index + 1:])
E             File "<string>", line 1
E               moin.log:138 using logging configuration read from "/home/moint03/test/moin/logging.conf"
E                       ^
E           SyntaxError: invalid syntax

/home/moint03/test/moin/src/moin/cli/_tests/__init__.py:53: SyntaxError
```

bylsmad:
@UlrichB22 I will fix the issue with MOINLOGGINGCONF, filed it as #1409

roland-ruedenauer:
@UlrichB22 This ticket is rather old and might be closed. We should open new issues if we see a need to improve the CLI tests.

Issue #1378, opened 2023-03-30
UnicodeEncodeError from call to print in cli


bylsmad:
windows terminal uses cp1252, if `print` is called with a utf-8 string containing characters which are not compatible with cp1252 an exception is raised, for example in `moin load-help -n en` if there is an itemname with non-ascii characters

replacing all calls to `print` with `logging.info` should resolve the issue

bylsmad:
fix will be ready for review once #1389 is merged, as I started the branch for the fix from there in order to have tests for the fix.

was able to resolve the issue with `sys.stdout.reconfigure(encoding="utf-8")` leaving the calls to `print` in place

PR #1389, opened 2023-04-03
tests for cli, #1387

bylsmad:
implementing tests for:
help
create-instance
index-create
load-help
item-get
item-put
index-dump

removed newline at end of file for *.data files
caused failure in item-get; item-put test

TODO: add tests for encoding issues, see #1378
TODO: create tests for other commands


src/moin/cli/_tests/test_modify_item.py
```diff
@@ -0,0 +1,100 @@
...
+        meta_cat_expected = json.load(f)
+    validate_meta(meta_cat_expected, meta_cat)
+    with open('cat.data', 'rb') as f:
+        cat_bytes = f.read()
+    with open(moin_dir / 'src' / 'moin' / 'help' / 'common' / 'cat.jpg.data', 'rb') as f:
+        cat_bytes_expected = f.read()
+    assert cat_bytes_expected == cat_bytes
+
+
+def test_item_put(index_create):
```
bylsmad, line 66: this one really should be using index_create2 to insure that it doesn't mess with test_dump_help

_test_artifacts/README.rst
```diff
@@ -0,0 +1,6 @@
+_test_artifacts directory
+========================
+
+this directory contains artifacts from tests
+
+any files dropped here will be uploaded in github action
```
bylsmad, line 6: looks like I missed the change to ci.yml which makes this statement true, since these tests don't leave anything behind in _test_artifacts it is not needed, will make the ci.yml change in another branch where there are actual files left behind to archive

UlrichB22:
Thanks for adding tests to cli. Some thoughts about the code:

- new python modules should get a header with copyright according to [src/moin/_template.py](https://github.com/moinwiki/moin/blob/master/src/moin/_template.py). Instead of `MoinMoin:YourNameHere` I commonly use `MoinMoin project` 
- Is it possible to keep the *.data files as they are? IMO every text file should have a linefeed at the end. What was the problem?
- Can we avoid having a new directory _test_artifacts in the source code as it seems to be created anyway when it is missing? This would help to keep the code simple and clean.
- I want to remove `-i -s` options on `moin index-create`, they are already obsolete. Can you try without?

I didn't test yet, this will need some time.

bylsmad:
the issue with the ending newlines was an item in storage which does not have a trailing newline would raise this exception if you item-get then item-put the resulting files:

`ValueError: computed data size (18) does not match data size declared in metadata (16)`

this issue is now resolved and the ending newlines are added back to the help/en files

I've addressed all the other issues you raised in your review, except I've left the non-english item in the help-en directory as it is the cause of the test failure in #1378 ... will move it in that branch

RogerHaase:
I am getting 4 file not found errors running tests on Windows.

https://gist.github.com/RogerHaase/921008302d20d30eec2764811ee15023

bylsmad:
@RogerHaase please retest, fix for windows pushed https://github.com/moinwiki/moin/pull/1389/commits/9351fa6c227fca526b6afc62ac4a11e05353533d

bylsmad:
@RogerHaase there will be a merge conflict once #1391 is merged ... we both fixed the issue with the trailing line endings created by  item-get in slightly different ways...I have a rebased version of my branch ready to go with fixes for the test, so let me know once #1391 is merged and I will push my rebased branch

UlrichB22:
Please also merge changes to index.py and modify_item.py from #1388. Thanks.

bylsmad:
@RogerHaase update: I investigated the line endings in the repo by setting my gitconfig `core.autocrlf false` and cloning your fork.  I found that even in the help branch both the help .data files as well as the .py files are stored in git with \n.  I've added comment on #1391 I am going to hold off on making the line ending changes in this branch

@UlrichB22 I just rebased this branch to the latest master which includes #1392 and #1388 cli tests are still passing

RogerHaase:
Running tests on windows now down to 1 error

https://gist.github.com/RogerHaase/547f111e971cc6ba5fe73824c5aef0d2

bylsmad:
@RogerHaase you found two bugs, one in the test one in cli  see https://github.com/moinwiki/moin/pull/1389/commits/50e7bacf3de46a69ceb3fc2e1a1b0833d94533d8

RogerHaase:
Thanks for helping!

Issue #1390, opened 2023-04-05
line endings on text help files should be \r\n


RogerHaase:
The text help files in /src/moin/help/en have \n line endings under version control. The size attribute in the .meta file is wrong, always higher than the actual file size. The sha1 attribute may be wrong, haven't checked.

If `moin dump-help -n en` or `moin item-get ...` is run on windows, then the output has \r\n line endings. The size attribute in the .meta file is correct.

`moin item-put ...`  changes the line endings of wiki text data files to \r\n, because html textareas have \r\n line endings. The meta file has no line endings.

moin dump-help and moin item-get should force \r\n line endings, and the help pages under version control should have \r\n line endings.

In theory this change should not impact unix based developers updating help files. The usual sequence would be to load help files, edit help files, dump help files, commit the changes, push.

bylsmad:
@RogerHaase I don't believe this branch has done what you intended.  my guess is you have your gitconfig set up with `core.autocrlf true` which replaces \r\n  with \n when you add a file.  This setting is a good thing IMHO because it makes for ease of working with folks like myself who use Linux or MacOS for development, I have mine set to `core.autocrlf input`.  (see the repo compare which shows the .data files are unchanged https://github.com/moinwiki/moin/compare/master...RogerHaase:moin:help)

the options as I see them:

1. leave the files with \n in git and drop this branch...on your system they will check out with  \r\n on my system they will check out with \n.  #1389 fixes the issue with the conflict in the size attribute on both Windows and Linux systems, please see the latest version of test_modify_item in that branch.  test_modify_item contains a test which reproduced the lineending at end of file issue before the fix was introduced in that branch.  note that the size and sha1 are calculated after the line ending conversion in moin.cli.maint.modify_item._PutItem so on Linux the size and sha1 appear to be wrong if I calculate them using the files on disc, but they are seen as correct when I item-put
2. add a .gitattributes file to the repo which  specifies that .data are binary files and redo this branch.  We will still need the fix in #1389 in order to handle item-get followed by item-put for a text item in storage which does not have a line ending at the end, so I'm not sure what is gained by this? see https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings?platform=windows

Either way, I don't see anything gained by forcing \r\n in the .meta files as python handles the line endings cross-platform seamlessly already for json files

I vote for option 1

RogerHaase:
[...]

Issue #1398, opened 2023-04-11
help-en meta files have invalid parentid attributes


RogerHaase:
These items have invalid parentid fields. StronaGłówna, MoinWikiMacros, mediawiki, html, and docbook.

PR #1400, opened 2023-04-11
line endings change for .data and .meta, #1390

bylsmad:
github action windows host uses core.autocrlf=true which causes the checked out files to have \r\n if they are text

converting to binary makes them \r\n everywhere, also makes the size in metadata match up

TIP: for git diff do `git diff --text` when you have changes


RogerHaase:
Before pushing changes to help files, `moin maint-reduce-revisions` must be run to avoid #1398. 

If you fix this I will kill PR 1399

bylsmad:
@RogerHaase new commit is added after running the moin maint-reduce-revisions, the parentids appear to be gone

RogerHaase:
One of the tools I use is Git for Windows that provides GIT GUI. For binary files GIT GUI says:

    Binary files a/src/moin/help/en/Home%2fsubitem.meta and b/src/moin/help/en/Home%2fsubitem.meta differ

which is not helpful in reviewing the changes before a commit > push.

TortoiseGit provides a diff that shows line by line changes for binary files but is more tedious to use. Are there similar issues with linux/mac tools?

A second concern is the .gitattributes file. Is "*.data binary" specific enough? Could it be "*/help/*.data binary" or similar?

Is there an alternative way to write the new tests without making the help .meta and ,data files binary?

RogerHaase:
My merge of PR1399 created above conflicts.

Issue #673, opened 2018-09-16
cleanup text types and quotes style for converter tests


ThomasWaldmann:
I've noticed that converter tests often use data like this (input, expected_output):

```
(u"......", '......')
```
Notice that the first element is unicode vs. the second being bytes.
Also, first one uses double quotes, second single quotes.

Considering our py3 future, all text-like objects should be unicode. (there's also another ticket about this)

Also, while cleaning that up, we can also unify the quote style.

This needs to be done in a separate cs with no other changes.

ThomasWaldmann:
For the in_out tests, it is also advisable to use this style:
```
(u"...",
 u"..."),
```
Then it is visually rather obvious that upper and lower string are same (or how they differ).

RogerHaase:
I tried using the double-quote style 

    (u"...",
     u"..."),

on test_moinwiki_in_out, but there are many instances where the output side of a test pair has embedded double quotes like

    ...class="red"...

So, if we want consistency, maybe we should favor single quotes like

    (u'...',
     u'...'),

as first choice, double-quotes as second choice, triple double-quotes as third choice, and triple single-quotes as last resort. And, whenever possible, the input and output pairs of a test case should use the same quoting style.

ThomasWaldmann:
Guess this can be closed.

If we want to auto-format some day, we could consider using `black`.

PR #1401, opened 2023-04-13
Converter test cleanup

JanHoefelmeyer:
Working towards https://github.com/moinwiki/moin/issues/673 .
I chose single quotes style unless single quotes were in the strings since this seemed to be the dominating method.

I did not change any input with triple quotes.

Otherwise I tried to satisfy https://github.com/moinwiki/moin/issues/673#issuecomment-421759156 as well.


RogerHaase:
Thanks for helping!

PR #1403, opened 2023-04-14
std in/out encoding issues in cli, fixes #1378

### LINKED ISSUE #1378: UnicodeEncodeError from call to print in cli

bylsmad:
updated test which handles line endings

removal on non-english content from help/en


RogerHaase:
looks OK, no issues running tests on windows

bylsmad:
this  line ending debacle took us some back and forth and a couple of merge conflicts, but I think we ended up with a very  clean solution - let the user decide what they want...nice working with you @RogerHaase

Issue #1402, opened 2023-04-13
moin save / moin load yields traceback


RogerHaase:
To reproduce create a new wiki, load help-en, take a backup, delete wiki, create an empty wiki, load backup.

Problem only occurs when help-en is loaded.

    moin load-help -n en
    moin save -f wiki/mybackup.moin -a true
    moin load -f wiki/mybackup.moin

8 help-en files are loaded, then:

    (moin-venv-python) C:\git\moin>moin load --file wiki/myback.moin
    2023-04-13 14:16:42,510 INFO moin.cli.maint.serialization:93 Load backup started
    Traceback (most recent call last):
      File "C:\GIT\moin-venv-python\Scripts\moin-script.py", line 33, in <module>
        sys.exit(load_entry_point('moin', 'console_scripts', 'moin')())
      File "C:\GIT\moin-venv-python\lib\site-packages\click\core.py", line 1130, in __call__
        return self.main(*args, **kwargs)
      File "C:\GIT\moin-venv-python\lib\site-packages\flask\cli.py", line 601, in main
        return super().main(*args, **kwargs)
      File "C:\GIT\moin-venv-python\lib\site-packages\click\core.py", line 1055, in main
        rv = self.invoke(ctx)
      File "C:\GIT\moin-venv-python\lib\site-packages\click\core.py", line 1657, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
      File "C:\GIT\moin-venv-python\lib\site-packages\click\core.py", line 1404, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "C:\GIT\moin-venv-python\lib\site-packages\click\core.py", line 760, in invoke
        return __callback(*args, **kwargs)
      File "C:\GIT\moin-venv-python\lib\site-packages\click\decorators.py", line 26, in new_func
        return f(get_current_context(), *args, **kwargs)
      File "C:\GIT\moin-venv-python\lib\site-packages\flask\cli.py", line 445, in decorator
        return __ctx.invoke(f, *args, **kwargs)
      File "C:\GIT\moin-venv-python\lib\site-packages\click\core.py", line 760, in invoke
        return __callback(*args, **kwargs)
      File "c:\git\moin\src\moin\cli\maint\serialization.py", line 95, in Deserialize
        deserialize(f, app.storage.backend, new_ns=new_ns, old_ns=old_ns, kill_ns=kill_ns)
[...]

PR #1418, opened 2023-04-23
test for cli load and save, #1387

bylsmad:
xfail tests for #1402
fix for help/welcome/Home.meta size
altering --path parameter on load-help
    now works same as dump-help
changing -a to flag for save
    appears this is how it is expected to work in quickinstall.py
fix for save with no -f option use stdout.buffer to output binary data

### DIFF
--- a/src/moin/cli/_tests/test_serialization.py
+++ b/src/moin/cli/_tests/test_serialization.py
@@ -1,0 +1,118 @@
+# Copyright: 2023 MoinMoin project
+# License: GNU GPL v2 (or any later version), see LICENSE.txt for details.
+
+"""
+MoinMoin - moin.cli.maint.serialization tests
+"""
+
+import json
+import os
+from pathlib import Path
+import pytest
+import sys
+
+from moin._tests import get_dirs
+from moin.cli._tests import run, assert_p_succcess, read_index_dump_latest_revs, getBackupPath
+from moin.constants.keys import NAME, MTIME, CONTENT, BACKENDNAME, NAMESPACE
+from moin import log
+
+logging = log.getLogger(__name__)
+
+
+def test_save_all(artifact_dir, save_all):
+    assert_p_succcess(save_all)
+    assert getBackupPath('backup.moin').exists()
+
+
+def load(restore_dir, backup_name, artifact_dir, args=None):
+    restore_dir.mkdir()
+    os.chdir(restore_dir)
+    try:
+        for command in (['moin', 'create-instance'],
+                        ['moin', 'index-create'],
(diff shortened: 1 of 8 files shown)


src/moin/cli/_tests/test_serialization.py
```diff
@@ -0,0 +1,118 @@
...
+            name = item[NAME][0]
+            content = item.pop(CONTENT)
+            metas[name] = item
+        assert set(expected_metas.keys()) == set(metas.keys())
+        for name, meta in metas.items():
+            expected_meta = expected_metas[name]
+            for k, v in meta.items():
+                if k in expected_meta:
+                    if k == MTIME:
+                        continue  # load replaces all mtimes
```
bylsmad, line 81: this seems a bit strange to me, is load from backup supposed to reset all the mtimes?

src/moin/cli/maint/modify_item.py
```diff
@@ -144,8 +145,10 @@ def LoadHelp(namespace, path_to_help):
     Load an entire help namespace from distribution source.
     """
     logging.info("Load help started")
-    abspath_to_here = os.path.dirname(os.path.abspath(__file__))
-    path_to_items = os.path.normpath(os.path.join(abspath_to_here, path_to_help, namespace))
+    if path_to_help is None:
+        abspath_to_here = os.path.dirname(os.path.abspath(__file__))
+        path_to_help = os.path.abspath(os.path.join(abspath_to_here, '../../help/'))
+    path_to_items = os.path.normpath(os.path.join(path_to_help, namespace))
```
bylsmad, line 151: making this work same as dump-help

src/moin/help/welcome/Home.meta
```diff
@@ -22,7 +22,7 @@
   "rev_number": 1,
   "revid": "be7556a9a778481c938ff656fe230c9d",
   "sha1": "59f6c196b7ff2db96cc620154e7c9addf46c33a4",
-  "size": 307,
+  "size": 324,
```
bylsmad, line 25: this is the correct size

src/moin/cli/maint/modify_item.py
UlrichB22: Perhaps you can try the same logic using moin_help as in 
https://github.com/moinwiki/moin/blob/bbd6ce2e9526ac8852b2b7f89d97d7c7c82e3225/src/moin/cli/maint/modify_item.py#L216
For me, it's easier to understand. What do you think about it?

bylsmad: @UlrichB22 done, see https://github.com/moinwiki/moin/pull/1418/commits/51ce36dd344f43509028fb51837c92cdbe2d9e9b

PR #1429, opened 2023-05-01
trivial change to item sort in dump_html.py

RogerHaase:
avoids users/Home used as _Home.html rather than default Home


UlrichB22:
Thanks for fixing dump-help cli command.

PR #1447, opened 2023-05-11
fixes for maint-set-meta command, fixes #1446

### LINKED ISSUE #1446: maint-set-meta issues
* fails with stacktrace 
[set-meta-stacktrace.log](https://github.com/moinwiki/moin/files/11454858/set-meta-stacktrace.log)

* once that is fixed, the command does not create a new revision, instead it sets the metadata on all revisions

bylsmad:


src/moin/storage/middleware/indexing.py
```diff
@@ -1233,7 +1233,8 @@ def store_revision(self, meta, data, overwrite=False,
         backend = self.backend
         if not overwrite:
             revid = meta.get(REVID)
-            if revid is not None and revid in backend:
+            backend_name = dict(app.cfg.namespace_mapping)[meta.get(NAMESPACE, '')]
+            if revid is not None and (revid in backend or (backend_name, revid) in backend):
```
bylsmad, line 1237: while developing the test this presented itself as a bug

 when SetMeta was still lacking the `del meta[REVID]` I did not get the expected ValueError until making this change

src/moin/_tests/test_forms.py
```diff
@@ -57,7 +57,7 @@ def test_validjson():
 
     Does not apply to usersettings form.
     """
-    app.cfg.namespace_mapping = [('', 'default_backend'), ('ns1/', 'default_backend'), ('users/', 'other_backend')]
+    app.cfg.namespace_mapping = [('', 'default_backend'), ('ns1', 'default_backend'), ('users', 'other_backend')]
```
bylsmad, line 60: this change is required by the change to storage.middleware.indexing below

Issue #1440, opened 2023-05-07
Cleanup and update docs


UlrichB22:
This issue can be used to track documentation updates.

UlrichB22:
Local docs and readthedocs don't show logo and favicon. 

`make html` gives these warnings

```
WARNING: logo file 'Logo_MoinMoin.png' does not exist
WARNING: favicon file 'favicon.ico' does not exist
```
According to the Sphinx docu the parameter html_logo and html_favicon need to be relative to the conf.py location.

UlrichB22:
Issue can be reopened if required.

PR #1457, opened 2023-05-30
Update README and add link to readthedocs

UlrichB22:
Related to #1440.


UlrichB22:
The intent is to put readthedocs first for documentation as it is updated regularly. Please approve.

PR #1492, opened 2023-08-29
Add OIDC authentication using Flask-pyoidc.

openbrian:
Here's a PR to add OIDC authentication.  It works just fine with Keycloak.

It does not fit into your design pattern because your Moin assumes all authentication methods have a password.  OIDC does not.  It will NOT login the user in middleware.

Initially there's support for 1 identity provider, but with small changes you can add multiple IDPs.

This change was very hard to make given the architecture of the current system.  I left comments where Moin can become a more idiomatic Flask website.  Notably:

* getting the user from the session should not depend on the authentication mode
* authentication should happen in a view/controller, not middleware
* the session needs only 2 items: user id, session_token
* model classes should not be touching the session
* the indexer is a pain in the ass to use / it should return objects not dictionaries

These are noted in the codebase with my initials: BCD.  Feel free to ignore and delete them.

A web app should not be inventing it's own authentication and session management.  Consider using Flask-Login to handle this.  (Also consider using Flask-Security-Too.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,7 @@ setup_args = dict(
     # optional features and their list of requirements
     extras_require={
         # 'featurename': ["req1", "req2", ],
+        'Flask-pyoidc>=3.14.2',  # OIDC SSO support
         'pillow': ["pillow"],  # successor to PIL; used by image get for scaling/rotating/etc.;
                                # requires special libs/header to be installed before it can be compiled successfully
         'ldap': ["python-ldap"],  # used by ldap auth; requires special libs/header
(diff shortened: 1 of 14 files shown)


RogerHaase:
setup.py:111:17: E999 SyntaxError: invalid syntax

Also had to uncomment these lines in wikiconfig.py:

    OIDC_REDIRECT_URI = 'http://localhost:5000/oidc_redirect_uri'
    OIDC_ISSUER="https://localhost:8080/auth/realms/myfolks"
    OIDC_CLIENT_ID='example.com/wiki'  # best practice is to use domain name and path here
    OIDC_CLIENT_SECRET='get this from the idp administrator'
    OIDC_LOGOUT_URI='todo'

Will test more tomorrow.

openbrian:
Start up an open ID connect server on port 8080.  I suggest keycloak.  Create a realm.  Add a user, set a password.  Create a client, set a client id.  Set the redirect uri that matches your env.

openbrian:
Or use a commercial one like Okta or Auth0.

Issue #1500, opened 2023-09-05
Select content_types that are allowed for new item creation


UlrichB22:
Wiki owners should be able to select the content_types that are allowed for new item creation. 
There may be various reasons, e.g. to disallow the creation of video or image elements or to disable svgdraw.

UlrichB22:
The last PR reverts the changes from the first PR. 
Contenttypes are removed from content_registry if disabled or not enabled. 
The content_registry is only used when creating new items AFAIK, so it is still possible to load e.g. help-en with different contenttypes or update existing items with disabled contenttype.

PR #1501, opened 2023-09-05
WIP: add feature to enable selected content types

UlrichB22:
This is a first draft for #1500.

I am still looking for a solution to move the `ENABLED_CONTENT_TYPES` variable into the configuration. It seems to be a problem because content_type registration is done before configuration is available.

I look forward to your comments and advice.

### DIFF
--- a/src/moin/items/_tests/test_Content.py
+++ b/src/moin/items/_tests/test_Content.py
@@ -26,6 +26,8 @@ from functools import reduce
 class TestContent:
     """ Test for arbitrary content """
 
+    @pytest.mark.skipif(len(Content.enabled_content_types) > 0 and 'PNG' not in Content.enabled_content_types,
+                        reason="'PNG' content_type disabled")
     def testClassFinder(self):
         for contenttype, ExpectedClass in [
                 ('application/x-foobar', Binary),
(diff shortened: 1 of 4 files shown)


UlrichB22:
I will add skips for pytest for disabled content_types later to prevent failures.

UlrichB22:
- Changed the default behavior to allow all content_types. 
- Added pytest skips in items/_tests for disabled content_types. 

Scrapy tests will still fail with errors when some content_types like 'TAR' or 'TGZ' are disabled. Don't know how to easily fix.

RogerHaase:
Possibly an easy way is to let registration run as is and fix the calls to `modify_select_contenttype.html` in `items/__init__.py`. 

Place `ENABLED_CONTENT_TYPES` into the default configuration using the same formats as:

                               group_names=content_registry.group_names,
                               groups=content_registry.groups,

That would remove the unwanted contenttypes from the UI.

To prevent hackers from creating a url with unwanted contenttypes add a check before the save in   `items/__init.py__` `def modify` (near line 776).

Note: 
iirc the line in `items/__init__.py near line 1538:

        return render_template('modify_select_contenttype.html',

is active, and a similar line near 1260:

                return render_template('modify_select_contenttype.html',

 is currently unused, awaiting the return of tickets and blogs.

When I removed tickets and blogs from the UI in 2017, the easiest way was to duplicate some code in

    def _select_itemtype(self):

skipping the choice for itemtype as there is only one choice for now: default

PR #1511, opened 2023-09-20
markdown: use wikilink extension

wagner-intevation:
For easier internal wiki links, activate the wikilink extension https://github.com/neurobin/mdx_wikilink_plus Previously, an internal link had to be: `[OtherPage](OtherPage)` now, this works too `[[OtherPage]]` and `[[Page/Subpage]]`

The wikilink extension of python-markdown itself (https://python-markdown.github.io/extensions/wikilinks/) is minimal and does not handle links with slashes in their name, so no subpages.

The complexity of creating the internal wiki links was/is one of the blockers of our upgrade to Moin 2.
The chosen wikilinks extension is listed at https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions and did fit the purpose best. Unfortunately, upstream made clear multiple times that their wikilink extension will stay as it is and not be extended (e.g. here https://github.com/Python-Markdown/markdown/issues/217)

### DIFF
--- a/docs/user/markdown.rst
+++ b/docs/user/markdown.rst
@@ -165,6 +165,22 @@ Inline links use the form: ::
 **reST NOTE**: Links with title attributes and images as links are not supported in reST.
 The internal links above are broken.
 
+Wikilinks
+---------
+
+Wikilinks use the form: ::
+
+    [[PageName]]
+
+===========================================   ===============================================
+ **Markup**                                    **Result**
+===========================================   ===============================================
+ [[Page]]                                      `Page <http:Page>`_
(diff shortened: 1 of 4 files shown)


wagner-intevation:
pushed an update to issues found by tests

wagner-intevation:
rebased on master

RogerHaase:
Thanks for helping!

The rst docs would benefit from more examples:

    With explicit label: [[ link | label ]]  # only supported for links not images
    Image: [[image.ext]]  # supports: .png, .jpg, .jpeg or .gif
    Image alt text: [[image.ext|alt=alternate text]]

And the help-en/markdown#Linking also needs an update.

PR #1525, opened 2023-09-25
Improve modify layout

cintek:
In most forms the buttons to confirm or cancel are at the bottom because in western countries we read from top to bottom. Because of that this PR moves the buttons for confirmation, cancellation and preview to the bottom. It also changed the order of the buttons because usually a button on the right lets users do anything to go on and a button on the left lets them go back. It also has to do something with how we read (from left to right; the right site is the "future").

Only the button to recover the draft stays at the top since it should be near the message which tells you that you can recover a draft.

~~The next change is that I also moved the alerts to the bottom because the feedback has to be where the buttons for interaction are.~~ (This causes problems, see discussion below)

Furthermore, this PR removes the `General meta` header because comments are also meta data.

This PR affects at least the basic and the modernized theme.

improve-modify-layout

### DIFF
--- a/src/moin/static/css/common.css
+++ b/src/moin/static/css/common.css
@@ -579,12 +579,12 @@ li.moin-selected-groups { font-size: 1em; font-weight: bold; }
 .moin-form dt { clear: both; float: left; width: 25%; text-align: right; margin-top: .3em; padding-right: 1em; }
 .moin-form dt label.required:after { content: '*'; color: var(--primary); }
 .moin-form button,
-.moin-form input[type="submit"] { clear: both; display: block; margin: auto; }
+.moin-form input[type="submit"] { clear: both; display: block; }
 .moin-form dl  { margin-bottom: 2em; }
 .moin-float-fix { clear: both; padding-top: .75em; }
 .moin-form  input[type="submit"].moin-modify-submit,
 .moin-form .moin-load-draft,
-.moin-form .moin-cancel { clear: none; float: left; margin-right: 3em; margin-bottom: 1em; }
+.moin-form .moin-cancel { margin-bottom: 1em; }
 .moin-form .moin-load-draft { border: 1px solid var(--alert);
     background-color: var(--bg-message); }
 .moin-watermark { background-image: url("../../../static/img/draft.png");
@@ -592,6 +592,9 @@ li.moin-selected-groups { font-size: 1em; font-weight: bold; }
 
 .moin-form dd.moin-radio { line-height: 1.5em}
 .moin-form .moin-radio input { width: 1em; }
+
+.moin-form .submit-buttons{ display: flex; column-gap: 8pt; justify-content: end;
+    margin: 8pt 0; } 
 
 #options dd { width: 10%; }
 #options dt { width: 60%; max-width: 40em; }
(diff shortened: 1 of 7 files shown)


RogerHaase:
Moving the "You may recover your draft ..." flash message to the bottom makes it less useful in that some/most editors would not realize a draft was available until they finish re-editing and scroll to the bottom to click the save button.

You could also delete the `General meta` title. Comment, Summary, and Tags are all meta data.

cintek:
> Moving the "You may recover your draft ..." flash message to the bottom makes it less useful in that some/most editors would not realize a draft was available until they finish re-editing and scroll to the bottom to click the save button.

In this case it really makes sense to show the flash on top of the page. I will have to take a look how it's possible to use the flash in two different places: On at the top and one where the buttons are.
 
> You could also delete the `General meta` title. Comment, Summary, and Tags are all meta data.

Alright

cintek:
I'm not sure what the best way is to solve this. Here is an idea:

We could use a HTML element at the top and at the bottom (where the buttons are) with class `moin-flash`. At the top we would show only messages like "You may recover your draft ..." by using if states. At the bottom we would show the rest of the messages, also by using if states.

This would only work if the text was not translated at that point but I'm not sure if that is the case.

A disadvantage is that at everytime someone changes the text of the message this part has also to be updated.

This example is for illustration of the idea:
```
<div id="top">
    <div class="moin-flash">
        {% for category, msg in get_flashed_messages(with_categories=true) %}
            {% if msg == "You may recover your draft ..." %}
            <p class="moin-flash moin-flash-{{ category }}">{{ msg }}</p>
            {% endif %}
        {% endfor %}
    </div>
</div>

<div id="bottom">
    <div class="moin-flash">
        {% for category, msg in get_flashed_messages(with_categories=true) %}
            {% if msg != "You may recover your draft ..." %}
            <p class="moin-flash moin-flash-{{ category }}">{{ msg }}</p>
            {% endif %}
        {% endfor %}
    </div>
</div>
```

UlrichB22:
> This would only work if the text was not translated at that point but I'm not sure if that is the case.
> 
> A disadvantage is that at everytime someone changes the text of the message this part has also to be updated.

You can try the categories feature of flashing, see https://tedboy.github.io/flask/patterns/flashing.html#flashing-with-categories. When you change the category to 'info-top' you can filter it in the template, e.g.:

```
<p class="moin-flash moin-flash-info">Flash messages:</p>
{% for category, msg in get_flashed_messages(with_categories=true) %}
    {% if category == "info-top" %}
        <p class="moin-flash moin-flash-info">{{ msg }}</p>
    {% endif %}
{% endfor %}
```
 
By the way, I am preparing a change regarding translations, which could also change some text phrases. IMO, using the msg text in an if statement is not an option.

cintek:
Sorry for the huge delay. I had much to do in our projects.

Thank you @UlrichB22 for your answer. I decided to let the flash stay where it is for now. As @RogerHaase explained it is no good idea to show the flash to recover the draft at the bottom. And most times a page is freshly loaded before you see a message so it's better to keep the flash at the top.

I will edit the description of the PR to reflect the changes that are now done by it.

BTW: You could even [filter the categories](https://tedboy.github.io/flask/patterns/flashing.html#filtering-flash-messages) which would make your solution even shorter, @UlrichB22 ;)

PR #1526, opened 2023-09-25
docs: fix markdown table example

wagner-intevation:
the first table example (markup description) was wrong.

Also makes the explanation of the non-supported table alignment in rst clearer in what it means for this docs section

### DIFF
--- a/docs/user/markdown.rst
+++ b/docs/user/markdown.rst
@@ -533,11 +533,11 @@ As shown in the second table below, use of outside borders and neat alignment of
 
 **Markup**: ::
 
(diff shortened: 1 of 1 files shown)


RogerHaase:
Typo in:

    **reST NOTE**: reST does not support cell alignment, 
    therefore the last example shown does not the 
    resulting alignment.

maybe:

    **reST NOTE**: reST does not support cell alignment, 
    therefore the last example shown above does not reflect the 
    resulting alignment.

I would leave the start/end "|" characters in the first example because (currently) a markdown to markdown conversion yields start/end "|" characters for all tables.

Please resubmit with corrections.

Issue #1467, opened 2023-07-07
drop support for TwickiDraw and AnyWikiDraw


RogerHaase:
Because TwickiDraw and AnyWikiDraw are written in Java and all modern browsers have dropped support for Java due to security concerns, there is no point in trying to support these obsolete drawing tools.

Attempts to create a new item results in an error:

    NOTE: You need a Java enabled browser to edit the drawing.

UlrichB22:
I will work on this issue if it's OK.

RogerHaase:
Great, I have not worked on this issue

PR #1529, opened 2023-09-26
remove TwickiDraw and AnyWikiDrawi support

UlrichB22:
Related to #1467. Please check ...

Is it ok to use .svgdraw as default extension in [wikiutil.py](https://github.com/UlrichB22/moin/blob/80593f672ac04e8b3503777d624ed47cd5ddb03c/src/moin/wikiutil.py#L199)? 
Is any "backward compatibility" still needed?

Be aware that existing local wikiconfig.py need to be adapted (remove 'twikidraw_moin' and 'anywikidraw'). The best option is to run a diff between `src/moin/config/wikiconfig.py` and the local copy. How can we inform users who created a wiki before?

Please advise.


RogerHaase:
Best way to inform wiki admins is moin-devel@python.org and moin-user@python.org mailing lists, and #moin channel. Per my light testing, the change to wikiconfig.py is a nice-to-have rather than a need. So maybe defer the notification unless you think it is necessary.

I will update https://moin-20.readthedocs.io/en/latest/admin/install.html#installation-for-developers to mention wikiconfig.py diff and `m extras`.

It is ok to use .svgdraw as default extension in wikiutil.py, no backward compatibility required.

Your changes look OK, merge when you are ready.

PR #1556, opened 2023-10-19
Raise werkzeug to 3.0 and remove werkzeug.urls usage

### LINKED ISSUE #1496: 'werkzeug.urls.url_quote_plus' is deprecated
'werkzeug.urls.url_quote_plus' is deprecated and will be removed in Werkzeug 2.4. Use 'urllib.parse.quote_plus' instead.

'werkzeug.urls.url_quote' is deprecated and will be removed in Werkzeug 3.0. Use 'urllib.parse.quote' instead.

Also werkzeug.urls.url_unquote

UlrichB22:
Fixes #1496.

### DIFF
--- a/src/moin/auth/__init__.py
+++ b/src/moin/auth/__init__.py
@@ -3,6 +3,7 @@
 # Copyright: MoinMoin:FrankieChow, MoinMoin:NirSoffer
 # Copyright: 2005-2012 MoinMoin:ThomasWaldmann
 # Copyright: 2007      MoinMoin:JohannesBerg
+# Copyright: 2023      MoinMoin project
 # License: GNU GPL v2 (or any later version), see LICENSE.txt for details.
 
 """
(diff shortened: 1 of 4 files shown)


src/moin/auth/__init__.py
```diff
@@ -3,6 +3,7 @@
 # Copyright: MoinMoin:FrankieChow, MoinMoin:NirSoffer
 # Copyright: 2005-2012 MoinMoin:ThomasWaldmann
 # Copyright: 2007      MoinMoin:JohannesBerg
+# Copyright: 2023      MoinMoin project
```
ThomasWaldmann, line 6: IANAL, but I doubt it works like that.

UlrichB22: With this entry I just want to document that the module is still maintained. 
An interesting view on copyright notices in open source can be found here:
https://opensource.com/article/20/10/copyright-notices-open-source-software

RogerHaase: I guess the issue is whether "MoinMoin project" qualifies as an owner under US copyright law and whether that matters in 2023. IANAL either, but my understanding is that a work can be registered in the US using a pseudonym. If you sue someone you have to prove you used the pseudonym. As GIT is tracking all the changes in detail "MoinMoin project" works for me.

PR #1587, opened 2024-02-02
fixes #1533 `dump-help -n en `updates all meta files

### LINKED ISSUE #1533: "moin dump-help -n en" updates all meta files
The mtime field in the meta file is updated even though the corresponding data file is not changed.

The order of entries within externallinks, itemlinks, and itemtransclusions is random so these fields may be updated.

Update `Updating the Help docs` within development.rst when fixed.

RogerHaase:


RogerHaase:
dk why those 4 month old commits are shown above. Seems harmless but will redo hoping for less confusing result

Issue #1585, opened 2024-02-01
Missing macros


UlrichB22:
There are some macros in moin1.9 that are not yet implemented. I found these in an existing moin1.9 wiki:

- Action
- FullSearch
- GoTo
- Navigation
- ShowSmileys
- SystemAdmin
- TitleIndex

I will add a PR to log missing macros in the migration with `moin import19`.

UlrichB22:
I would like to create a simple SlideShow Macro as replacement for `<<Action(SlideShow)>>`.
The  idea is to have single page slide shows as described in [HelpOnSlideShows](https://moinmo.in/HelpOnSlideShows).
I would be pleased to get some hints and suggestions for the implementation.

RogerHaase:
Actions went away in moin2 before I started working on the project. Looked, but could not find a relevant discussion in the old Google Summer of Code logs. I found this history of enhancements: http://moinmo.in/RichardFlieger/Ideas/SlideShow

I assume SlideShow should be implemented as a macro copying code from the 1.9 `action/SlideShow.py`. Do not implement the <<Action..>> macro.

UlrichB22:
Regarding **Navigation macro**:

According to http://moinmo.in/HelpOnMacros the Navigation macro in 1.9 has these features:

 - `<<Navigation(children[,maxdepth])>>`  macro to list all of a pages' children.
 - `<<Navigation(siblings[,maxdepth])>>`  quick navigation to siblings of a page.
 - `<<Navigation(slideshow)>>`  creates a link to toggle between slide show and wiki mode, and a link to start the slideshow. 
 - `<<Navigation(slides)>>`  displays first/prev/next/last links.

The ItemList macro can be used to list links for children and siblings. 

`<<Navigation(children)>>` can be replaced with `<<ItemList(display="ChildName")>>`

The differences in the output are:
- the result is shown as list and not in one line
- the childnames are not preceeded with a "/"

We just need to add `<<Navigation(children)>>` to the macro migration.

The SlideShow marco in moin2 only supports slideshows based on a single wiki item and the above mentioned links are shown by default. IMO we don't need to do anything.

Do you have any suggestions?

UlrichB22:
Regarding **FullSearch macro** and categories

The FullSearch macro in moin 1.9 is often used in conjunction with categories, see http://moinmo.in/HelpOnCategories

Example:
`<<FullSearch(category:CategorySamples)>>`

[...]

PR #1601, opened 2024-02-12
Add macro TitleIndex

UlrichB22:
related to #1585.

Documentation, help-en updates and tests will follow.


UlrichB22:
will create a new one ;-)

Issue #1597, opened 2024-02-11
import19 cannot import a moin1.9 wiki farm


ClaasTeichmann:
I tried to import my running moin1.9 wikifarm (14 wikis and 18GB of data alltogether). I followed the the instructions from the website (https://moin-20.readthedocs.io/en/latest/admin/upgrade.html) and from the code.
I couldn't find any documentation for importing a wikifarm (sorry if I overlooked something!).

## Present setup (moin1.9):
File/directory structure:

    ./farm/farmconfig.py
    ./farm/wiki1.py
    ./farm/wiki2.py
    ./farm/ ....
    ./farm/common/user/
    ./farm/wiki1/data/pages/
    ./farm/wiki1/data/cache/
    ./farm/wiki1/ ....
    ./farm/wiki2/data/pages/
    ./farm/wiki2/data/cache/
    ./farm/wiki2/ ....

## (expected) Future setup
I understood that former wikis of a wikifarm translate into namespaces of a moin2 wiki (please correct me if I am wrong here).
So I would expect the following structure:

    ./farm/wikiconfig.py
    ./farm/index
    ./farm/preview
    ./farm/sql
    ./farm/data
    ./farm/data/default
    ./farm/data/help-common
    ./farm/data/help-en
    ./farm/data/userprofiles
    ./farm/data/users
    ./farm/data/wiki1
    ./farm/data/wiki2
    ./farm/data/ ...

## Problem
It seems that I can convert each of my wikis one by one, but the pages are assigned to the __default__ namespace and not the individual ones, which I already configured in the wikiconfig.py

I can only import one wiki of the wikifarm at once:

    moin import19 -d ./moin19_wikifarm/wiki1/data

### 1. The import19 could detect the wikifarm automatically and assign the namespaces accordingly:

    moin import19 -d ./moin19_wikifarm

### 2. There could be the option to which namespace the wiki should be converted:

    moin import19 --namespace=wiki1 -d ./moin19_wikifarm/wiki1/data
    moin import19 --namespace=wiki2 -d ./moin19_wikifarm/wiki2/data
    ...

I am happy to write the instructions or do further testing if this helps the moin2 development.

UlrichB22:
I think solution 2 should be the easiest. I will check it out, but it will take some time.
Perhaps you can continue in the meantime with one of your largest or most important wikis to test whether there are any showstoppers. Thanks.

UlrichB22:
[...]

PR #1609, opened 2024-02-19
import19: add target namespace parameter

UlrichB22:
... and fix itemlink migration for subitems (e.g. for users namespace)

Related to #1597.

### DIFF
--- a/src/moin/cli/migration/moin19/import19.py
+++ b/src/moin/cli/migration/moin19/import19.py
@@ -32,7 +32,7 @@ from moin.app import create_app
 from moin.constants.keys import *  # noqa
 from moin.constants.contenttypes import CONTENTTYPE_USER, CHARSET19, CONTENTTYPE_MARKUP_OUT
 from moin.constants.itemtypes import ITEMTYPE_DEFAULT
-from moin.constants.namespaces import NAMESPACE_DEFAULT, NAMESPACE_USERPROFILES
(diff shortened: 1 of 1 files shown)


ClaasTeichmann:
I tested import19 with my wikifarm and found several issues, which I am listing here:

## Missing directories
I have to create the missing directories:

    wiki01/data
    wiki01/meta

before converting.

## User directories

User directories are stored in moin19 in

    common/user
    
They are expected by the import routine in
wiki01/data/user

Solution:

    cd wiki01/data/
    ln -s ../../common/user .

Alternatively, it could be good to provide an additional parameter to specify the user-directory.

## Groups

In moin19 each wiki in the farm had its own EditorGroup, AdminGroup, etc.
Now, to separate the EditorGroups, it would be good to convert the EditorGroup to Wiki01EditorGroup.
The wikiconfig.py has to be adapted accordingly.

Do I understand it correctly that different wikis with different users are separated now by using different namespaces?

## Missing Link

This, e.g., pdf cannot be found:

    http://127.0.0.1/namespace01/Page01/Attachment01.pdf

It can be found here, thus the namespace information is missing:

    http://127.0.0.1/Page01/Attachment01.pdf

Mind that the page is created correctly here:

    http://127.0.0.1/namespace01/Page01

## EditorGroup doesn't exist

The page

    http://127.0.0.1/EditorGroup

doesn't exist and cannot be created.

Issue #1596, opened 2024-02-10
add list of forbidden_user_names to constants/misc.py


RogerHaase:
Within wikiconfig.py there are powerful user and group names that should be changed by a wiki admin prior to enabling public access. Per #1555 there is a concern that a careless wiki admin may accidently allow a hostile user an opportunity to self register a powerful user name.

As noted in #1561, user names may be created using the special groups names All, Known, and Trusted. This appears to be harmless, just odd.

One way to force wiki admins to change wikiconfig.py is to disallow the use of the example names by checking a list of forbidden names during account / group creation.

    forbidden_user_names = ['YourSuperUserName', 
        'SomeOtherSuperUserName',
        'YourSuperEditor'
        'All',
        'Known',
        'Trusted', ]
    forbidden_group_names = ['MyTrustedEditorGroup', ]

UlrichB22:
IMO the check should be case-insensitive. Human errors may occur if wiki admins do not realize different spellings (e.g. "SuperUserName" and "SuperUsername").

ThomasWaldmann:
Hmm, instead of adding more code to fix potential security issues caused by the default wikiconfig, the default could be also changed to not give these high privileges - e.g. the acl definitions could be commented and an additional comment above these could read the users must first define those user names before activating the ACLs for them.

That's a general best practice btw and not limited to the wikiconfig. E.g. you could just put some acl onto a wiki page mentioning a username that does not exist yet. Better always first create the user, then use it in ACLs.

ThomasWaldmann:
About "All", "Known", "Trusted": is it the case that currently a user named like this can:
- be created
- can get the permissions meant for the group with same name without being in that group?

That would be a bug I guess.

ThomasWaldmann:
@UlrichB22 please never introduce case-insensitivity anywhere. it usually just causes lots of troubles and is not as easy as you might think.

RogerHaase:
Using:

        NAMESPACE_DEFAULT: dict(before='RogerHaase,Trusted:read,write,create,destroy,admin',
                                default='All:read,write',
                                after='',
                                hierarchic=True, ),

then self-registering as a user named Trusted; the logged-in Trusted user gets default read,write permissions.

Cannot find a way where users named Known or Trusted get upgraded to permissions of special groups.

----
Will try for updated wikiconfig.py while keeping it simple for desktop users. 

Will ignore the users named All, Known, Trusted issue as it seems harmless.

PR #1627, opened 2024-02-24
fix wikiconfig.py instructions for setting up single user wiki #1596

RogerHaase:


src/moin/config/wikiconfig.py
```diff
@@ -18,10 +18,10 @@
     registration_only_by_superuser = True
 to:
     registration_only_by_superuser = False
-and change several lines containing:
-    dict(before='YOUR-SUPER-EDITOR:read,write,create,destroy,admin',
+and change four lines containing:
```
ThomasWaldmann, line 21: Do you really want to count the lines that need changing every time?

RogerHaase: Just trying to make it simple and explicit for desktop wiki users. If the number changes, I will put it back to "several".

Issue #1237, opened 2022-05-04
Replace create_simple_mapping with create_mapping in wikiconfig.py


RogerHaase:
Having two ways of creating storage mapping requires having to document both ways and when to use one way rather than the other. The addition of help for editors in separate namespaces makes create_simple_mapping less useful as the number of languages increase.

While create_mapping is more verbose, it is easy to understand and can be added to wikiconfig.py with common defaults.

Issue #563, opened 2017-04-28
userprofiles directory is created but not used


ThomasWaldmann:
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/563) by RogerHaase (Bitbucket: [RogerHaase](https://bitbucket.org/RogerHaase), GitHub: [RogerHaase](https://github.com/RogerHaase)).**

----------------------------------------

If a wiki is using the default settings in wikiconfig.py and creates a new wiki (./m sample, ./m new-wiki), then under the directory <wiki root>/wiki/data/, sub-directories __default__ and __userprofiles__ are created. However, the __userprofiles__ directory is not used.

This is caused by a bug in storage/\_\_init__.py create_simple_mapping.

The lines:

    backends = {
        BACKEND_DEFAULT: None,
        BACKEND_USERPROFILES: None,
    }

create the userprofiles directory, but the " + '/'" here:

    namespaces = {
        NAMESPACE_DEFAULT: BACKEND_DEFAULT,
        NAMESPACE_USERPROFILES + '/': BACKEND_USERPROFILES,
    }

results in all userprofiles being added to the default directory.

Also there is a similar issue in:

    acls = {
        NAMESPACE_USERPROFILES + '/': userprofiles_acl,
        NAMESPACE_DEFAULT: default_acl,
    }

One problem with fixing this is that if there is a wikiconfig_editme.py overriding these defaults, the wiki admin may not notice the change.

An easy solution should be to do a __/.m backup xxx__ followed by a __./m restore xxx__. The restore will magically restore any userprofiles in the default directory to the userprofiles directory. However, this only works one time as the load procedure only restores the default directory. See #564.

https://moin-20.readthedocs.io/en/latest/admin/configure.html#namespaces should be updated to better explain this feature.

PR #1650, opened 2024-03-15
remove unwanted foo`/` from wikiconfig.py examples; fixes #1649

### LINKED ISSUE #1649: wikiconfig.py: "trailing / below causes foo to be stored in default backend"
From wikiconfig.py line ~156:

        # define custom namespaces if desired, trailing / below causes foo to be stored in default backend
        # 'foo/': 'default',

The trailing / is wrong, but worked OK until the fix for #1446 was applied where line ~1226

    if revid is not None and revid in backend:

was replaced by:

            backend_name = dict(app.cfg.namespace_mapping)[meta.get(NAMESPACE, '')]
            if revid is not None and (revid in backend or (backend_name, revid) in backend):

The new code fails because meta.get(NAMESPACE) = 'foo' and app.cfg.namespace_mapping has a key "foo/" but not a key "foo".

    -- snip --
      File "C:\GIT\moin\src\moin\items\__init__.py", line 1059, in _save
        fqname, new_meta = storage_item.store_revision(meta, data,
      File "C:\GIT\moin\src\moin\storage\middleware\protecting.py", line 412, in store_revision
        rev = self.item.store_revision(meta, data, overwrite=overwrite, return_rev=return_rev, fqname=fqname, **kw)
      File "C:\GIT\moin\src\moin\storage\middleware\indexing.py", line 1262, in store_revision
        backend_name = dict(app.cfg.namespace_mapping)[meta.get(NAMESPACE, '')]
    KeyError: 'foo'

The fix is to remove the trailing /

RogerHaase:


src/moin/config/wikiconfig.py
```diff
@@ -150,11 +151,11 @@ class Config(DefaultConfig):
         NAMESPACE_USERS: 'users',
         NAMESPACE_USERPROFILES: 'userprofiles',
         # namespaces for editor help files are optional, if unwanted delete here and in backends and acls
-        'help-common': 'help-common',  # contains media files used by other language helps
-        'help-en': 'help-en',  # replace this with help-de, help-ru, help-pt_BR etc.
-        # define custom namespaces if desired, trailing / below causes foo to be stored in default backend
-        # 'foo/': 'default',
```
ThomasWaldmann, line 156: The idea behind that was that pages starting with `foo/` (== subpages of foo) are stored in that namespace.

If you remove the slash, page `foobar` and `fool` will also be stored into that namespace.

Except if there was a code change that auto-appends a slash. Is that the case?

RogerHaase: The trailing / was from `/storage/__init__.py create_simple_mapping` #563   when wikiconfig.py used create_simple_mapping. The trailing / is no longer in the docs https://moin-20.readthedocs.io/en/latest/admin/configure.html#create-simple-mapping but lived on in code comments.

wikiconfig.py now uses `create_mapping` per #1237 which allows admins to create custom namespaces.

    namespaces = {
        # maps namespace name -> backend name
        # first, configure the required, standard namespaces:
        NAMESPACE_DEFAULT: 'default',
        - snip -
        # some additional custom namespaces stored in default backend:
        'foo': 'default',
        'bar': 'default',
        # custom namespace with a backend
        'baz': 'baz',
    }

There were 2 code changes, trailing / was removed from create_simple_mapping, and wikiconfig.py changed to use create_mapping.

PR #1658, opened 2024-04-04
flask 2.3 update related changes

ThomasWaldmann:


ThomasWaldmann:
I checked the Flask 2.3.x change log about other stuff that was removed, but seems like the sendfile stuff is the only removed thing moin used.

PR #1663, opened 2024-04-04
run pyupgrade plus a few manual fixes

ThomasWaldmann:


ThomasWaldmann:
This is a quite big change and I didn't review the pyupgrade changes, but I have used it in other projects before and it worked flawlessly yet.

PR #1668, opened 2024-04-04
readthedocs: build on ubuntu 22.04 using python 3.10

ThomasWaldmann:
Python 3.10 is the system python on 22.04.


ThomasWaldmann:
works!

PR #1682, opened 2024-04-11
readthedocs does not provide access to /docs/examples/*; fixes #1578

### LINKED ISSUE #1578: readthedocs does not provide access to /docs/examples/*
In /docs/examples/config, /docs/examples/deployment, /docs/examples/password-reset there are several example files with names not ending in .rst.

There are some references to these files within https://moin-20.readthedocs.io but afaik there is no way to view these files. 

Wiki admins installing moinmoin from pypi have no means of viewing these files because the docs directory is not included in the distribution.

Preferred solution is for readthedocs to have working links to the files.

RogerHaase:
some of the files moved to contrib/* are not referenced in the docs

moin.fcgi, moin.wsgi, and test.wsgi were moved but not changed, black reformatted after push


docs/admin/backup.rst
```diff
@@ -16,7 +16,7 @@ Selective Backup
 ================
 If you want a backup of MoinMoin and your data, then backup the following:
 
-* your data
+* your data, ususlly everything under wiki/
```
ThomasWaldmann, line 19: typo

docs/admin/configure.rst
```diff
@@ -564,7 +564,8 @@ username (like with german umlauts or accented characters). If moin does not
 crash (log a Unicode Error), you have likely found the correct coding.
 
 For users configuring GivenAuth on Apache, an example virtual host configuration
-file is included with MoinMoin in `docs/examples/deployment/moin-http-basic-auth.conf`.
+is included at `/contrib/deployment/moin-http-basic-auth.conf`
```
ThomasWaldmann, line 567: no leading "/"?

docs/admin/configure.rst
```diff
@@ -1570,13 +1571,13 @@ https://docs.python.org/3/library/logging.config.html#configuration-file-format
 
 
 There are also some logging configurations in the
-`docs/examples/config/logging/` directory.
+`contrib/logging/` directory.
```
ThomasWaldmann, line 1574: better here!

docs/admin/configure.rst
ThomasWaldmann: (also applies at some other places)

docs/admin/configure.rst
RogerHaase: Thanks for corrections, will fix

PR #1693, opened 2024-05-18
Support for SQLAlchemy 2.0

### LINKED ISSUE #1661: update moin.storage.stores.sqla to use sqlalchemy 2.0+
sqlalchemy >= 2.0 needs some changes, e.g. working based on a connection.

guess applying the required changes to that store would be very easy for somebody who is used to working with sqlalchemy.

our tests can be invoked like `pytest -k stores` to check if that works fine.

davestgermain:
Fixes #1661


ThomasWaldmann (approved):
Thanks for the PR!

LGTM!

Issue #1699, opened 2024-06-13
'Error: metadata validation failed' when adding a link with starting '+' sign 


UlrichB22:
If you create a new item called Test and add a link that points to the metadata of this item:  `[[+meta/Test]]`,
you can save and view the item, the link works fine. If you try to modify the item and click on "ok" to save it, you will receive the message:

`Error: metadata validation failed, invalid field value(s) =`

The CLI command to check metadata doesn't find an issue:

```
$ moin maint-validate-metadata -v -a
0 items with invalid metadata found
```

Same problem exists with any other url starting with a '+' sign. Deleting the item leads to a HTTP 500 and traceback in the logs. The workaround to fix the problem was manually removing the itemlinks entry in the metadata on the filesystem layer :frowning_face: .

The issue is similar to #1577.

UlrichB22:
The validation of itemlinks is using wikiname_validator and name_validator and they don't allow the usage of views like `+meta/<itemname>`:

https://github.com/moinwiki/moin/blob/d913e8703f4e6740bdb374fd77e6a32fa56c7ce6/src/moin/storage/middleware/validation.py#L398

Should we create a new itemlink_validator to solve the issue? Please advise.

ThomasWaldmann:
Didn't we use the "+" as prefix for actions?

RogerHaase:
Agree, we need an itemlink_validator to resolve this issue.

Another problem is the links have a class of moin-nonexistent:

    <a class="moin-nonexistent" href="/+Highlight/Test2">+Highlight/Test2</a>

PR #1703, opened 2024-06-26
Fix moin-nonexistent class in itemlinks starting with '+'

UlrichB22:
Related to #1699.

It is hard to solve the problem with class non-existent. I added a hack in lines 205f of `converters/link.py` to show a possible solution for view '+meta', for example. IMO adding a dictionary with view_names and endpoint-function-names seems to be too complex. 

Any ideas on how to improve this? Need help.

### DIFF
--- a/src/moin/app.py
+++ b/src/moin/app.py
@@ -30,7 +30,7 @@ from whoosh.index import EmptyIndexError
 from moin.utils import monkeypatch  # noqa
 from moin.utils.clock import Clock
 from moin import auth, user, config
-from moin.constants.misc import ANON
+from moin.constants.misc import ANON, VALID_ITEMLINK_VIEWS
 from moin.i18n import i18n_init
 from moin.themes import setup_jinja_env, themed_error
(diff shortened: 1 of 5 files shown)


RogerHaase (commented):
Agree this is the best way to fix the problem. 

If constants/misc.py remains as:

    VALID_ITEMLINK_VIEWS = ["+meta", "+history", "+download", "+highlight"]

maybe add a comment that +delete, +convert etc. will show as non-existent links,
else just add them all.

RogerHaase (approved):
Agree this is the best way to fix the problem. 

If constants/misc.py remains as:

    VALID_ITEMLINK_VIEWS = ["+meta", "+history", "+download", "+highlight"]

maybe add a comment that +delete, +convert etc. will show as non-existent links,
else just add them all.

UlrichB22:
Fixes #1699.

In create_app a function has been added that retrieves the relationship between view name and endpoint to avoid a manually maintained dictionary with redundant data. I hope this is OK.

Issue #1707, opened 2024-07-04
User Settings > Personal allows users to create invalid names


RogerHaase:
##### 1. Self-registration or command line registration checks the user name for invalid characters such as `/`, `,`, `:`, etc.

But a logged-in user can change his name to an invalid name such as `Joe/Doe` by using the User Settings > Personnel dialog.

##### 3. Related: missing feature? Moin 1.9.x had configure options to prevent users from changing their name or email: 

    # +++ these are suggested changes to the user preferences form if the  external_cookie  is the only auth method
    user_form_disable = ['name', 'aliasname', 'email',] # don't let the user change these, but show them:
    user_form_remove = ['aliasname', 'password', 'password2', 'css_url', 'logout', 'create', 'account_sendmail','jid'] # remove completely:

Ignore 3 above; moved to #1712.

##### 4. Related: per docs:

    Display-Name
     The display name can be used to override your username, 
     so you will still log in using your username
     but your display name will be displayed to other users 
     and in your history page.

The Display-Name is not shown in Global History, Item History, nor the footer. Instead the (first) username is displayed.

UlrichB22:
Flush error messages are only visible for a very short period of time and disappear without being clicked. 
I had this problem already before this change in my environment.

PR #1710, opened 2024-07-12
User Settings > Personal allows creation of invalid display names #1707

RogerHaase:


RogerHaase:
This is wrong approach - should use Flatland validation.

PR #1744, opened 2024-08-27
Change font to Poppins

rivercory:
I changed the font to Poppins to improve readability!
Please review my PR and give me your feedback :)


UlrichB22:
In the EU we have a General Data Protection Regulation (GDPR) and because of this it is problematic to include fonts from an external resource. Wiki Admins may use these fonts in a custom CSS setup but we should not use them in our code base.

rivercory:
@UlrichB22 If I make the font local, will that solve the problem?

Issue #1781, opened 2024-10-17
Page trail: checking the existence of items slows down the response time for larger wikis


UlrichB22:
I tested an imported wiki with 2400 items and about 12000 revisions. Each whoosh index query lasts about 120 ms. 
At the start of my tests the browser history was empty, I just showed TestItem01, TestItem02, ... TestItem05 and again TestItem01. 
The following server output was produced with additional clock timers:

```
DEBUG 2024-10-17 21:55:54,261 moin.utils.clock:48 timer total(0): 1564.11ms /TestItem01
DEBUG 2024-10-17 21:55:58,699 moin.utils.clock:48 timer total(0): 1602.45ms /TestItem02
DEBUG 2024-10-17 21:56:03,658 moin.utils.clock:48 timer total(0): 1795.20ms /TestItem03
DEBUG 2024-10-17 21:56:08,248 moin.utils.clock:48 timer total(0): 2165.31ms /TestItem04
DEBUG 2024-10-17 21:56:13,020 moin.utils.clock:48 timer total(0): 2321.03ms /TestItem05
DEBUG 2024-10-17 21:56:17,832 moin.utils.clock:48 timer total(0): 2376.22ms /TestItem01
```
With 5 entries in the page trail, the response time increased by 800 ms. Two additional index queries are used for each item.

The relevant parts in the code are:

https://github.com/moinwiki/moin/blob/807bfe10d56cf37e878667e0363849ab1f7e3df4/src/moin/themes/__init__.py#L352

https://github.com/moinwiki/moin/blob/807bfe10d56cf37e878667e0363849ab1f7e3df4/src/moin/themes/__init__.py#L373

UlrichB22:
IMO there are two solutions to regain performance:

1. add the exists status while showing an item to the page trail. If an item is deleted in parallel by someone else the status may not be accurate.
2. do not add non-existing items to the page trail and remove the checks.

When I try to open an item by entering a non-existing itemname in the browser URL the create dialog is show. At this moment the item is already added to the page trail even if I leave the dialog without adding the item. Not sure if this is useful.

RogerHaase:
120 ms seems like a long time just to see if an item exists. I keep thinking that somewhere in the processing below storage.get_item there is some code that opens the data file of every item in the page trail and includes it as part of the Item object.. But I have been unable to find it. It would be more comforting if the procedure name were storage.get_item_meta.

Agree that having non-existent items in the page trail is not very useful, and your solutions above.

[...]

Issue #1725, opened 2024-08-13
Slow performance: up to 10s per page


sebix:
The load time of one of our wikis has significantly decreased over time as more and more pages were added. The wiki still has only 69 pages, but the load times vary between 2s and 10s, which significantly reduces the user experience.

On the same server, we have an even smaller wiki with the same code base and only 13 pages, with load times of 0.7-1.7s.

Because of this correlation, we guess that the performance degradation is related to the number of pages.

Are there any performance-related things to improve in the settings?
Are the any logging/debugging options to investigate the cause?

We are currently on version 5d33343dc617c706fd5d9d9fe398f2493f53e286

UlrichB22:
Can you try to run a `'moin index-optimize'` command when nobody is updating the wiki? In #1310 this fixed a performance issue. 

Before and after running that command you can list the index files with `'ls -l wiki/index'` to compare the size and number of index files.

sebix:
Thank you for the tip.

The number of files decreased from 20 to 6 and the total size of the directory from 62M to 56M.
The load times are now indeed below 2s, which isn't great but a big relief.

I guess we need to run that command on a regular basis.

ThomasWaldmann:
@sebix but you hopefully don't run moin as cgi?

With a persistent python/moin2 wsgi process, 2s sounds way too much for a normal/simple page.

Of course, on special pages and pages with expensive macros, load time depends very much on what is actually done there.

sebix:
> @sebix but you hopefully don't run moin as cgi?
> 
> With a persistent python/moin2 wsgi process, 2s sounds way too much for a normal/simple page.

The setup uses apache2 with WSGI.

> Of course, on special pages and pages with expensive macros, load time depends very much on what is actually done there.

The mentioned load times were all for simple pages with markdown or text only, without macros. `/+index/` and `/+history/` actually have the same load times (1.5s-2s).
Markdown seems to have a very little effect of about 0.2s or so.
Calling non-existent pages is quicker by about 1s: 0.8s.
Calling, for example, `/Home` or `/+index/Home` directly from localhost to exclude any possible delays of TLS, reverse proxies, etc., the load time is consistently 1.7s.

bernhardreiter:
> try to run a 'moin index-optimize' command when nobody is updating the wiki?

Is there an easy way to disable that people can update the wiki? Or: What happens if somebody updates it while the command runs?

UlrichB22:
[...]

PR #1782, opened 2024-10-18
Deactivate exists checks for page trail and ignore non-existent items

UlrichB22:
Related to #1781.

The cleanup of the code related to the “exists” field will be done later.


UlrichB22:
Also related to #1725.

UlrichB22:
The aliases have been moved to the page trail, as each check for aliases previously required an index query.

Issue #1783, opened 2024-10-22
Import19 optimizations


UlrichB22:
This issue is intended to document some optimizations of the import19 migration program.

1. the import19 command should have 'procs' and 'limitmb'-options to be used for the index-build step to improve performance.
2. tracebacks in the `'drop and rebuild index'` step should be handled
3. when parent items for sub-items are missing (`'These items have orphan sub-items'`-message in the +index view), they should be created automatically.
4. ...

UlrichB22:
Another optimization:

 4. add option --latest-rev-only

With the `--latest-rev-only` option, you can omit the history of the pages and only import the latest revision of each item into the new wiki. This is particularly useful for testing the migration to moin2.

PR #1784, opened 2024-10-22
import19: Add procs and limitmb options to increase performance

UlrichB22:
related to #1783.


RogerHaase:
Item creation with ~1,000 items ~5,000 revisions seems faster!

PR #1795, opened 2024-11-10
update intermap.txt removing dead links, spam: fixes #1793

### LINKED ISSUE #1793: update intermap.txt
1. sort by wikiname making it easier update by testing each link
2. test each link, delete anything that doesn't resemble a public wiki

RogerHaase:


UlrichB22:
Thanks for updating intermap.txt.

I am wondering about these two entries:

```
Google http://www.google.com/search?q=
GoogleGroups http://groups.google.com/groups?q=
```

They are not public wikis. When I try to open groups, I get a login page.

We should also switch to https if possible.

RogerHaase:
Thanks for tip. Changed most http: to https:

Left the google entries and a few other non-wikis as is thinking that at least one user may be using the links.

Issue #1820, opened 2024-12-08
Bandit static security code analysis: fix findings and add bandit to github actions


UlrichB22:
This issue collects several fixes for 'bandit' findings.

I would also like to add Bandit to the Github actions as described in 
https://github.com/marketplace/actions/bandit-scan or
https://github.com/marketplace/actions/bandit-by-pycqa.

UlrichB22:
```
hashlib: Use of weak MD5 hash for security. Consider usedforsecurity=False
Test ID: B324
Severity: HIGH
Confidence: HIGH
CWE: [CWE-327](https://cwe.mitre.org/data/definitions/327.html)
File: [src/moin/utils/crypto.py](file:///home/hu/Documents/Source/moinwiki/moin2_git/moin/src/moin/utils/crypto.py)
Line number: 120
More info: https://bandit.readthedocs.io/en/1.8.0/plugins/b324_hashlib.html

119	    """
120	    return hashlib.md5(repr(kw).encode()).hexdigest()
121
```

```
hashlib: Use of weak SHA1 hash for security. Consider usedforsecurity=False
Test ID: B324
Severity: HIGH
Confidence: HIGH
CWE: [CWE-327](https://cwe.mitre.org/data/definitions/327.html)
File: [src/moin/utils/plugins.py](file:///home/hu/Documents/Source/moinwiki/moin2_git/moin/src/moin/utils/plugins.py)
Line number: 205
More info: https://bandit.readthedocs.io/en/1.8.0/plugins/b324_hashlib.html

204	        assert isinstance(pdir, str)
205	        modname = "moin_p_{}".format(hashlib.new("sha1", pdir.encode()).hexdigest())
206	        if modname not in sys.modules:
```

IMO these functions are used to generate unique names, so the addition `'usedforsecurity=False'` is appropriate here.

UlrichB22:
```
try_except_pass: Try, Except, Pass detected.
Test ID: B110
Severity: LOW
Confidence: HIGH
CWE: [CWE-703](https://cwe.mitre.org/data/definitions/703.html)
[src/moin/utils/send_file.py](file:///home/hu/Documents/Source/moinwiki/moin2_git/moin/src/moin/utils/send_file.py)
Line number: 141
More info: https://bandit.readthedocs.io/en/1.8.0/plugins/b110_try_except_pass.html

140	                fsize = file.tell()  # tell position
141	            except Exception:
142	                pass
[...]

PR #1822, opened 2024-12-09
Rework of the send_file exception handling

UlrichB22:
Related to #1820.

### DIFF
--- a/src/moin/utils/send_file.py
+++ b/src/moin/utils/send_file.py
@@ -31,6 +31,11 @@ from werkzeug.wsgi import wrap_file
 from flask import current_app, request
 
 
+from moin import log
+
+logging = log.getLogger(__name__)
+
+
 def encode_rfc2231(value, coding="UTF-8", lang=""):
     """
     Encode a value according to RFC2231/5987.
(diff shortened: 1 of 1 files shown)


src/moin/utils/send_file.py
```diff
@@ -127,24 +132,21 @@ def send_file(
...
         # and tell methods, but they just raise some exception (e.g. UnsupportedOperation)
         # instead of really doing what they are supposed to do (or just be missing).
         try:
             file.seek(0, 2)  # seek to EOF
-            try:
-                fsize = file.tell()  # tell position
-            except Exception:
-                pass
+            fsize = file.tell()  # tell position
             file.seek(0, 0)  # seek to start of file
```
ThomasWaldmann, line 146: if it succeeds with seek, fails with tell, it will not seek back to start of file.

but, it will log a warning AND CONTINUE PROCESSING (and malfunction, because it is at EOF position).

UlrichB22: Added a final seek.

ThomasWaldmann (approved):
LGTM.

PR #1823, opened 2024-12-09
Remove obsolete SubProcess.py

UlrichB22:
Related to #1820.


src/moin/utils/SubProcess.py
```diff
@@ -1,209 +0,0 @@
-# Copyright: 2003-2005 by Peter Astrand <astrand@lysator.liu.se>
-# Copyright: 2009,2010 by Thomas Waldmann (modifications)
-# License: Python license (http://www.python.org/2.4/license)
-
-"""
-Enhanced subprocess.Popen subclass, supporting:
-    * .communicate() with timeout
```
ThomasWaldmann, line 7: that's present in stdlib since python 3.3.

PR #1830, opened 2025-01-24
Show blank page after password reset

### LINKED ISSUE #1805: "Access denied" error after password reset if /Home is not readable for anonymous
From a user feedback.
Setup: internal wiki, no pages readable for anonymous users. The user account was created by the admin (as user registration is disabled)

When the user resets the password (initial password), the user first receives an e-mail with the reset link. On that page is a form with the pre-filled token, name, and new password. After submitting the form, the page shows:

In a blue info block on the top: "Your password has been changed, you can log in now."
The heading "Access Denied" (the biggest element on the screen)
In normal text: "You are not allowed to access this resource."

The innocent user sees the most prominent element on the screen, "Access denied," and may interpret this as an error.
But the error text is not related to the reset action, but on the page's ACL.

Possible solutions:
1. Display the text "you can log in" even bigger than "Access denied"
2. Recommend never requiring authentication to view `/Home.`
3. Instead of showing `/Home` after the reset, show a separate page and only the success message
4. Login to the user automatically after the reset (it was authenticated with the token anyway).

UlrichB22:
...
 - sorted robots.txt
 - removed TODOs. IMO the handling of a next page is obsolete if a blank page is always displayed.

Fixes #1805.

### DIFF
--- a/src/moin/apps/frontend/views.py
+++ b/src/moin/apps/frontend/views.py
@@ -138,41 +138,41 @@ def robots():
         """\
 User-agent: *
 Crawl-delay: 20
-Disallow: /+convert/
-Disallow: /+dom/
-Disallow: /+download/
-Disallow: /+modify/
-Disallow: /+content/
-Disallow: /+delete/
+Disallow: /+admin/
 Disallow: /+ajaxdelete/
 Disallow: /+ajaxdestroy/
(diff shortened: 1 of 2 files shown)


sebix:
Thank you!

PR #1834, opened 2025-01-30
Add bandit scan to github workflow

UlrichB22:
Used a customized copy of the action.yml from https://github.com/PyCQA/bandit-action.

The use of bandit-action was not possible because it uses Python 3.9. Reading the configuration in pyproject.toml requires an additional installation step for toml. This tool is integrated in Python since 3.11. If bandit-action raises the Python version, we can switch to bandit-action later.

Bandit results do not abort the workflow, it is terminated with success. The results are imported into the github security area using a sarif file. This area is protected and not publicly accessible. Tested with branch name 'test-github-action'.

image


UlrichB22:
Fixes #1820.

PR #1840, opened 2025-02-20
fix CSP warning, remove inline style on moin-options-for-javascript #…

RogerHaase:
…1816


UlrichB22:
I have tested the fix and it works as expected. Maybe you can just add spaces and a semicolon to be compliant with the existing code style?

`#moin-options-for-javascript { display: none; }`

RogerHaase:
thanks for spotting rookie error

Issue #1847, opened 2025-03-03
Duplicate namespace in generated discussion link when viewing the sitemap of a wiki page


roland-ruedenauer:
When viewing the page `http://localhost:5000/+sitemap/users/roland` in the browser and hover
the discussion link I find the 'users' namespace duplicated in the generated url:
`http://localhost:5000/users/users/roland/Discussion`.

The problematic link is generated in this template:
https://github.com/moinwiki/moin/blob/aa6f024b81fdc71f0253c9eb9636b7af8c32e119/src/moin/templates/itemviews.html#L122

This line should probably corrected into:

`{%- set supp_name = '%s/%s' % (fqname, sub_item_name) %}`

Note that the discussion link generated looks ok when just viewing the page (not on the sitemap view).

Moin version: 2.0.0b2

UlrichB22:
@roland-ruedenauer, thanks for reporting this bug. Do you want to create a PR? You solution looks good.

PR #1850, opened 2025-03-03
Fix duplicate namespace in discussion link (#1847)

roland-ruedenauer:
Fix the issue described in #1847


UlrichB22:
Thanks for supporting moin!

PR #1854, opened 2025-03-06
Fix download link for binary content (missing namespace)

roland-ruedenauer:
The converter for binary content (moin.converters.everything) in moin version 2.0.0b2 ignores the item namespace when creating the dom content. The result is an invalid download link when viewing the subitem.

Downloads are working when using the fully qualified item name.

Question in this context: shouldn't the link text created for downloading be localized? Currently it seems to hard-coded to "Download \<item-name\>."

### DIFF
--- a/src/moin/converters/_tests/test_everything.py
+++ b/src/moin/converters/_tests/test_everything.py
@@ -1,0 +1,51 @@
+# Copyright: 2025 MoinMoin
+# License: GNU GPL v2 (or any later version), see LICENSE.txt for details.
+
+"""
+MoinMoin - Tests for moin.converters.everything
+"""
+
+import pytest
+
+from . import serialize, XMLNS_RE
+
+from moin.constants.keys import CONTENTTYPE, ITEMTYPE, REV_NUMBER
+from moin.converters.everything import Converter
+from moin.items.content import Content
+
+from moin.utils.interwiki import split_fqname
+
+from moin.items import Item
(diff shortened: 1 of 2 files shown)


RogerHaase:
I think you missed this change to the Development docs https://moin-20.readthedocs.io/en/latest/devel/development.html

install the pre-commit hook:

    pre-commit install  # pre-commit is used for code linting / auto-format

and the commit to master is hung up (and rather hidden).

If you click on the red X above and then click on Details you will see Black is complaining about some code formatting.

I think a way to fix this is to:

run the "pre-commit install" command
make a minor change to everthing.py (add a trailing blank), then try to commit
Black will complain about the code format, and correct it automatically
do de-stage, stage cycle
commit, push

UlrichB22:
You are right about the localization. There is already a translation for the word “Download”. Perhaps we can use that.

roland-ruedenauer:
Sorry, I obviously didn't carefully read your contributor guidance documents and missed the pre-commit hooks.

I've pushed the re-formatted changes (using `pre-commit run` is the simplest way to trigger the hooks as I found out).
There is now a new translation for the download link text. The already existing "Download" didn't seem suitable to me.

I also tried to update the i18n messages as documented, but found the latest release of pybabel adds lots of new comment lines containing `#, python-brace-format` and skipped this step.

RogerHaase:
Thanks for helping!

PR #1859, opened 2025-03-08
Fix undefined name 'revs' in diff view

roland-ruedenauer:
The PR fixes a `NameError: name 'revs' is not defined` when attempting to view a diff against a bookmarked history entry.

### DIFF
--- a/src/moin/apps/frontend/views.py
+++ b/src/moin/apps/frontend/views.py
@@ -2766,8 +2766,8 @@ def diff(item_name):
                 rev1 = revid
                 break
         else:
-            rev1 = revs[-1][1]  # if we didn't find a rev, we just take oldest rev we have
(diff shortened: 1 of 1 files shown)


UlrichB22:
Can you please explain the steps how to reproduce the error?

roland-ruedenauer:
To reproduce the error, try this:
* visit the history of a namespace (here "users")
* click on the "Set bookmark" link next to the date of an old revision
* revisit the namespace history
* click on a diff button in the first column (marked red in screenthost - hover it and you get "show diff based on bookmark")
* experience an "Internal Server Error"

UlrichB22:
This PR fixes the Traceback.

But I'm not sure about the logic. I tested an item with 11 revisions, created a fresh bookmark and got a diff between revision 4 and 11.

PR #1865, opened 2025-03-14
Remove test code from python wheel

roland-ruedenauer:
As the title says, this PR changes the packinging of moin.
Files below any of the "_tests" folder will no longer be part of the created python wheel.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,24 +84,23 @@ ldap = ["python-ldap"]  # used by ldap auth; requires special libs/header
 moin = "moin.cli:cli"
 
 [tool.setuptools]
-# See also the MANIFEST.in file.
-# We want to install all the files in the package directories...
-include-package-data = true
+include-package-data = false
 
 [tool.setuptools.packages.find]
 where = ["src"]
-exclude = ["_tests"]
+exclude = ["*._tests", "*._tests.*"]
+namespaces = true
 
 [tool.setuptools.package-data]
-"moin.translations" = ["MoinMoin.pot", "*.po", ]
-"moin.static" = ["*", ]
-"moin.themes.modernized" = ["*", ]
-"moin.themes.basic" = ["*", ]
-"moin.themes.topside" = ["*", ]
-"moin.themes.topside_cms" = ["*", ]
-"moin.templates" = ["*.html", "*.xml", ]
-"moin.apps.admin.templates" = ["*.html", ]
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Discuss: maybe we rather want to be able to run the tests, even if moin is installed from a sdist or wheel?

roland-ruedenauer:
I should have added, that it was attempted before to not have the test code in the python wheel.
Meaning this PR is kind of packaging fix.

https://github.com/moinwiki/moin/blob/6bf364a06730301da771bef3d83fefe7b3809e42/pyproject.toml#L93

And with the PR test code should still be present in the created sdist.

But of course, it might be wanted now to still have the test code also in the Python wheel.

UlrichB22:
By excluding the test modules, we can keep the moin package clean and focused on the wiki admins who install and update a moin wiki.

The tests are run multiple times when a PR is created and merged into the master branch, the test results are archived through the GitHub workflow. I see no reason to repeat the tests for an “old” package version.

RogerHaase:
Googling gives mixed results, and yields "including tests are recommended, but not required".

https://packaging.python.org/en/latest/tutorials/packaging-projects/ seems to suggest including tests:

-----
# Creating the package files
You will now add files that are used to prepare the project for distribution. When you’re done, the project structure will look like this:

packaging_tutorial/
├── LICENSE
├── pyproject.toml
├── README.md
├── src/
│   └── example_package_YOUR_USERNAME_HERE/
│       ├── __init__.py
│       └── example.py
└── tests/
## Creating a test directory
tests/ is a placeholder for test files. Leave it empty for now.

-----

I vote to continue including tests, as some odd wikiconfig file may cause a failure.

roland-ruedenauer:
Thanks for commenting.

I have to admit, I did not expect someone could want to have test code and data be part of a wheel package. The installer most of the time will install wheels in some site-packages folder and all you normally would want to do now imo is run the installed application or use the library. And if I wanted to run tests for a package, then I would either download the sdist or more likely directly clone the source repository.

Btw, the reason I created the PR is quite simple: I built a docker container for moin2 where I installed the beta-2 wheel package from pypi. Then I opened a shell in the running container to verify it's content and was surprised to see all of the _tests folders were present in the site-packages folder.

Anyways, feel free to close the PR if you like to keep the packaging as it is.

UlrichB22:
We still need some cleanup of the existing pyproject.toml to remove the wrong exclude line. This should be done in a new PR.

PR #1871, opened 2025-03-18
Exclude UI tests in the bandit configuration

roland-ruedenauer:
Changing the pre-commit config is required in addition because files passed by the commit hook to bandit via the command line are always checked despite being excluded by the configuration.

See: [Skip tests folder on pre-commit](https://github.com/PyCQA/bandit/issues/912)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,9 @@ repos:
   - id: bandit
     args: ["-c", "pyproject.toml"]
     additional_dependencies: ["bandit[toml]"]
+    exclude: |
+      (?x)(
+        ^quickinstall.py|
+        /_tests/|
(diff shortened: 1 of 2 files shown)


UlrichB22:
The exclude of directories for pre-commit is not working for me. I copied quickinstall.py to _ui_tests/test_bandit.py,
run `'pre-commit install --install-hooks'`. Tested different syntax variants without success.

The pyproject.toml change is ok, exclude of quickinstall.py is also ok.

roland-ruedenauer:
@UlrichB22 Thank you for testing this. I corrected the exclude regex in the pre-commit config. It works for me (on my Linux machine).

UlrichB22:
Looks good now, thanks for supporting MoinWiki.

PR #1877, opened 2025-03-21
Add 3rd minimal whoosh index for performance

UlrichB22:
Related to #1725.

After updating to this version the indexes need to be dropped and rebuild:

```
moin index-destroy
moin index-create
moin index-build
```

This will take some time for large wikis.

The index-build subcommand will create a new third index called LATEST_META.  This index will be much smaller than LATEST_REVS and will only contain the usual metadata fields, but no content or content ngrams. 

Many queries to check the existence of an item or to check ACL rights only require a few metadata fields. Queries against this small index are very fast and improve response times for large wikis.

The new index is called LATEST_META and the parameter to use the new index in various methods or functions is called “short”. 

If you revert to a version prior to this change, you should delete and recreate the indexes. Advanced users can remove the new index files (*latest_meta*) in the wiki/index directory instead.


src/moin/apps/frontend/views.py
```diff
@@ -572,6 +572,9 @@ def wrapper(item_name, rev):
             abort(404, item_name)
         if add_trail:
             flaskg.user.add_trail(item_name, aliases=item.meta.revision.fqnames)
+        if rev == CURRENT:
+            """if view has been called with default rev=CURRENT we can avoid an index query in flash_if_item_deleted"""
+            item.is_current = CURRENT
```
ThomasWaldmann, line 577: Nitpick:

When using is_something as a name, the name suggests that the value is a boolean, indicating whether the object in question "is something" (or not).

But it looks like the value isn't boolean here.

RogerHaase:
A guess is you thought about removing content from the latest_revs indexes. Is it easier to create a 3rd index or did you do some benchmarking?

src/moin/storage/middleware/indexing.py
```diff
@@ -757,18 +808,21 @@ def update(self, tmp=False):
...
-                ix_revids = {doc[REVID] for doc in searcher.all_stored_fields()}
-            backend_latest_revids = {revid for name, revid in backend_latest_backends_revids}
-            upd_revids = backend_latest_revids - ix_revids
-            upd_revids = [(revids_backends[revid], revid) for revid in upd_revids]
-            self._modify_index(index_latest, self.schemas[LATEST_REVS], upd_revids, "update")
-            self._modify_index(index_latest, self.schemas[LATEST_REVS], del_revids, "delete")
-        finally:
-            index_latest.close()
+
+        # update LATEST_REVS and LATEST_IDX
```
ThomasWaldmann, line 812: Maybe rather tell here WHY we have the smaller index.

src/moin/cli/maint/index.py
```diff
@@ -13,7 +13,7 @@
 from flask.cli import FlaskGroup
 
 from moin.app import create_app, init_backends
-from moin.constants.keys import LATEST_REVS, ALL_REVS
```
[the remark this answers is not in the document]
UlrichB22: LATEST_META might be a good compromise. I will change it in the final version after getting some performance figures from other users.

UlrichB22:
> A guess is you thought about removing content from the latest_revs indexes. Is it easier to create a 3rd index or did you do some benchmarking?

Some time ago I tested removing the content, this will impact searching and search results. IMO the different NGRAMs have the biggest impact on the index size. This solution with a third index should not change the search functionality.

UlrichB22:
With the latest commit the recommended changes have been applied.
The new index is now called `'latest_meta'`.

Running `'moin run'` without recreating indexes will result in the following error message:

`Error: Wiki index 'latest_meta' missing. Please see https://github.com/moinwiki/moin/pull/1877`

src/moin/storage/middleware/indexing.py
UlrichB22: The explanation is now in lines 449-450:

```
    # Small index for the latest revisions, used for queries such as has_item, authorization checks and
    # the +index route. This index has no content or *NGRAMS, which improves query speed for large wikis
```

UlrichB22:
@ThomasWaldmann, @RogerHaase, may I ask you to review this PR. IMO this change is urgently needed for large wikis (e.g. python.org) to get reasonable response times. Thank you.

RogerHaase:
Sorry for the delay. Busy with other things, no time for moin.

For my wiki on windows with 900+ items, response times dropped from about 5 seconds to about 2 seconds. Very nice fix.

Issue #1838, opened 2025-02-12
Markdown parser crashes on emphasized text in deleted text in list item


sebix:
This markdown text:
```
 * <del>Deleted list item with _emphasized text_</del>
```

causes the following exception:

```
2025-02-12 16:54:11,193 ERROR moin.items.content:316 An exception happened in _render_data (error_id = 08f9c665-393d-4890-94d7-e67430dc4e7f ):
Traceback (most recent call last):
  File "/home/sebastianw/dev/moin/src/moin/items/content.py", line 304, in _render_data
    doc = self.internal_representation(preview=preview)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sebastianw/dev/moin/src/moin/utils/clock.py", line 65, in wrapper
    retval = f(*args, **kw)
             ^^^^^^^^^^^^^^
  File "/home/sebastianw/dev/moin/src/moin/items/content.py", line 260, in internal_representation
    doc = input_conv(preview or self.rev, self.contenttype, arguments=attributes)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sebastianw/dev/moin/src/moin/converters/markdown_in.py", line 632, in __call__
    self.convert_embedded_markup(converted)
  File "/home/sebastianw/dev/moin/src/moin/converters/markdown_in.py", line 540, in convert_embedded_markup
    self.convert_embedded_markup(child)
  File "/home/sebastianw/dev/moin/src/moin/converters/markdown_in.py", line 540, in convert_embedded_markup
    self.convert_embedded_markup(child)
  File "/home/sebastianw/dev/moin/src/moin/converters/markdown_in.py", line 540, in convert_embedded_markup
    self.convert_embedded_markup(child)
  File "/home/sebastianw/dev/moin/src/moin/converters/markdown_in.py", line 536, in convert_embedded_markup
    node[idx] = self.embedded_markup(child)  # child is immutable string, so must do node[idx]
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sebastianw/dev/moin/src/moin/converters/markdown_in.py", line 519, in embedded_markup
    p_text = html_in_converter(f"<p>{text}</p>")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1882, opened 2025-03-28
Fix misc issues found while testing moin

roland-ruedenauer:


roland-ruedenauer:
With these issues fixed entering the "problematic" markdown content in #1838 no longer results in a Oops! page.

Issue #1770, opened 2024-09-28
Improve Basic Theme


RogerHaase:
In Show view, a QuickLinks mouseover makes text nearly invisible.

In Meta view, attributes and values are centered in content area making them difficult to read. Other themes have attributes and values aligned on left. Making the Meta view a 2-column table would improve readability on all themes.

In Modify View, clicking on Edit ACL somehow disables the Edit Content and Edit Meta tabs. User must click browser refresh to regain control.

In Modify view, both Edit Meta and Edit ACL tabs have input fields for ACL.

On top banner near Settings/Logout, current user's name is displayed in red, whether user's home page exists or not. Link is to user's home page in users namespace.

The View Options, User Actions, and Item Actions panels are initially hidden. A bit of JS and browser local storage could set these to last use similar to 'Toggle font width' on Modify view.

roland-ruedenauer:
I'm currently working on this topic.

cintek:
Why not drop this theme completely and concentrate on the others? @RogerHaase your comments in https://github.com/moinwiki/moin/pull/1528 indicated already that many themes mean more maintenance so it would probably good to remove one of theme or move it to the theme market. I just took a look at the themes and I would say that the basic theme has the lowest usability of all themes.

You have found already some issues and there are more potential things that could be improved like the low contrast of some elements. And the collapsible element "View Options" contains only one item so there is no reason to hide it in the first place.

PR #1892, opened 2025-04-09
Render item metadata as a table

roland-ruedenauer:
Picking up the suggestion to improve readability made in #1770.


roland-ruedenauer:
Please check again

Issue #1621, opened 2024-02-23
quickinstall.py and "m" need to die


ThomasWaldmann:
These are nice convenience wrappers for developers and "git checkout users", but for a normal installation (like after pip install or after installing a moin linux dist package) I guess they won't work.

Also, these are VERY special to moin and feel alien even to users familiar with the usual python installation / usage workflows.

These 2 scripts do all sorts of stuff and depending on the target user group I think we need different ways to replace these scripts:
- for the developer-only oriented helpers, I guess we could use `tox` or even `make`. For another project I just created a `scripts/make.py` to do misc. stuff there (including a `clean` subcommand, replacing `python setup.py clean`). No stuff that a end users needs should be in there.
- for the moin admins / "users", I think we either should integrate into a `moin` (or `moin-admin`) script command (these will be provided by `pip install` and also by a linux dist package install).
- guess everything dealing with virtual envs should be removed from our helpers. That is outside of the scope that moin tools should cover and there are good tools for that, like e.g. `pipx`, or just the basic `python -m venv` python users are very familiar with. Pycharm, for example, also supports devs creating and activating venvs and trying to do that again in moin might rather collide with such stuff.

ReimarBauer:
When there is a pypi package we can use [grayskull](https://github.com/conda/grayskull) to create a conda-forge package. I also added there some of the xstatic packages already.

sfermigier:
Yes. I think once the virtualenv is set up (and there are tools for that: poetry, pdm/hatch, rye...), administrative commands should be available via subcommands of the `flask` CLI.

ThomasWaldmann:
Guess this should be fixed before release?

`m` and `quickinstall` were hacks addressed to developers, but they shouldn't be required for users - moin should behave and be usable like a normal python project.

PR #1898, opened 2025-04-16
.gitignore: mark some files as not to be ignored

roland-ruedenauer:


roland-ruedenauer:
The copies of wikiconfig,py and intermap.txt possibly created by `./m create-instance` should be ignored by git.
But the the 2 files existing in the source tree used as  *source* for the copy operation should for sure be tracked be git.

roland-ruedenauer:
> IMO each of those lines should be placed directly behind the corresponding ignore pattern

Currently you have a group of exclude patterns grouped below the comment `# files created by quickinstall.py`. Would you really place these new _don't ignore_ patterns within this group?

Maybe be more verbose and explain why not ignore some files? Finally there are other tools besides git, that evaluate the content of `.gitignore`.

UlrichB22:
At the moment this is fine. 

I was looking for the reason why we have these files in the root directory. They are copied in the first run of quickinstall.py, even if you put your wiki directory in a different location, which is my recommendation.

https://github.com/moinwiki/moin/blob/02dea301a32a655caa49cabd01fd7b0de070bc48/quickinstall.py#L721)

In this case, these files will become out of date over time. One step to get rid of quickinstall.py (see #1621) would be to remove all wiki maintenance commands, as they are also covered by moin CLI subcommands. Then we can clean up .gitignore.

PR #1903, opened 2025-04-24
Add support for scoped subitem search using > prefix

### LINKED ISSUE #1885: search an item's subitems
In apps/frontend/views.py there is:

    @frontend.route("/+search/<itemname:item_name>", methods=["GET", "POST"])
    @frontend.route("/+search", defaults=dict(item_name=""), methods=["GET", "POST"])
    def search(item_name):

with the explanation:

    "Currently" there is no theme generating the '/+search/<itemname:item_name>' link
    within Item Views. To access, users must key the query link into the browsers URL. The
    query result is filtered limiting the output to the target item, target subitems
    and sub-subitems..., and transclusions within those items.
    Example URL: http://127.0.0.1:8080/+search/OtherTextItems?q=moin

and more comments regarding searching transclusions of subitems:

            if show_transclusions:
                # XXX Search subitems and all transcluded items (even recursively),
                # still looks like a hack. Imaging you have "foo" on main page and
                # "bar" on transcluded one. Then you search for "foo AND bar".
                # Such stuff would only work if we expand transcluded items
                # at indexing time (and we currently don't).

Search currently has the Best Match feature:

### Redirect to best match

    If you know the target item name, start the search term with a back-slash character. 
    Only the names and namengram fields will be searched. If there is a hit, 
    the browser will be redirected to the highest scoring hit.

Searching an item's subitems could be implemented by adding a one-character prefix 
to the user generated query. e.g. `>mypage blue` would search all sub-items of `mypage` for `blue`.

mmheyer:
Implements feature request [#1885](https://github.com/moinwiki/moin/issues/1885) by enabling users to scope searches to a specific item and its subitems using a > prefix (e.g., >Lectures design patterns). Includes:
- Query parsing and scoping logic
- UI message update for scoped searches
- Unit tests for query parsing

RESOLVES #1885

### DIFF
--- a/src/moin/_tests/test_views.py
+++ b/src/moin/_tests/test_views.py
@@ -1,0 +1,27 @@
+"""
+    MoinMoin - moin.views Tests
+"""
+
+from moin.apps.frontend.views import parse_scoped_query
+
+
+class TestParseScopedQuery:
+    def test_parse_scoped_query_with_prefix_and_query(self):
+        scope, actual_query = parse_scoped_query(">Lectures design patterns")
+        assert scope == "Lectures"
+        assert actual_query == "design patterns"
+
+    def test_parse_scoped_query_with_prefix_only(self):
+        scope, actual_query = parse_scoped_query(">Lectures")
+        assert scope == "Lectures"
+        assert actual_query == ""
+
+    def test_parse_scoped_query_without_prefix(self):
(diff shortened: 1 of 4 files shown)


RogerHaase:
@mmheyer did you notice you have a failing test?

Found another search bug #1907. OK to fix this PR, work on 1907 later or leave for someone else.

PR #1906, opened 2025-04-30
pin xstatic packages to legacy versions

ThomasWaldmann:
I added some upper version limit for all xstatic* packages, so it uses the proven legacy versions.

The upper limit should be removed again after xstatic 2.0.0 is released and after all xstatic-* used by moin were modernized, released and tested with moin. Then the limits shall be adjusted so that the modernized version is the minimum requirement (to avoid any mixup of legacy and modernized packages).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,16 +55,16 @@ dependencies = [
     "pdfminer.six",  # pdf -> text/plain conversion
     "passlib >= 1.6.0",  # strong password hashing (1.6 needed for consteq)
     "sqlalchemy >= 2.0",  # used by sqla store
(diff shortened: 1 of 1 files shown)


UlrichB22:
Shouldn't we also raise some minimum versions if they are more than 2 or 3 years old? 
I've just checked the obvious ones:

```
XStatic >= 1.0.3  # 2023-05-04
XStatic-jQuery == 3.5.1.1  # 2020-11-08
XStatic-Pygments == 2.9.0.1  # 2021-05-13
```

ThomasWaldmann:
@UlrichB22 There is no need to do that now. After all modernized packages are available/released, it will need an update to these versions as minimum version anyway.

Issue #1680, opened 2024-04-10
Some ReStructuredText links broken in moin2.


gmilde:
It seems like "simple" links to different Wiki pages fail in text parts written in rST:

Working examples: links with targed specified in rST
(cf. https://moin-20.readthedocs.io/en/latest/user/rest.html#hyperlinks).

    Links that work
    =============

    This is a `hyperlink with included target <WikiLink>`_.

    This is a `hyperlink to an explicit target`_.

    .. _hyperlink to an explicit target: WikiLink

    This is an `anonymous hyperlink to an explicit target`_.

    __ WikiLink

    This is a link to an implicit target: `links that work`_.

All these links work in standard rST

Now the failing example: A link to a different WikiPage without explicit or embedded target.
This works in the "old" MoinMoin but fails in Moin2:

    Here is a link to a MoinMoin page named SecondPage_.

Details: 
> The reference would typically cause an unknown target error from the docutils parser. This is because there isn't a target in the document named "[SecondPage](https://moinmo.in/SecondPage)". However, with the [MoinMoin](https://moinmo.in/MoinMoin) parser, the "[SecondPage](https://moinmo.in/SecondPage)_" reference instead creates a link to a [MoinMoin](https://moinmo.in/MoinMoin) page named "[SecondPage](https://moinmo.in/SecondPage)".
 — https://moinmo.in/HelpOnParsers/ReStructuredText section "Unknown Targets"

It seems that a part of the underlying code was transferred to moin2: see the [visit_reference() function](
https://github.com/moinwiki/moin/blob/41267f832ee4b88610728384501d7b0b4c392c19/src/moin/converters/rst_in.py#L583)

However, this code is never reached, as Docutils checks for references without targets earlier (in a "transform") and creates a "system message" describing the problem (With Docutils and Sphinx, the system message is inserted into the output, in moin2 it seems to be shown in the log only.)

If continued support for simple WikiLinks_ is planned, moin2 would need to 

* define a "resolver" function and register it in the "unknown_references_resolvers" hook,
* adapt the abovelinked visit_reference() function to the [upcoming changes in Docutils DTD](https://docutils.sourceforge.io/RELEASE-NOTES.html#document-tree-docutils-dtd).

gmilde:
The patch to future-proof the `rst_in.py` reStructured Text input converter:

~~~ diff
--- /tmp/rst_in.py.bak	2025-05-07 21:43:33.229924365 +0200
[...]

PR #1921, opened 2025-05-21
Fix reference handling in reStructuredText.

gmilde:
cf. issue #1680

Fix `refuri` for internal hyperlinks:

IDs given to section headings tend to differ from the IDs generated by Docutils. Explicit targets keep the ID generated by Docutils. Find out whether the target node is a section and try to generate the ID value from its heading. 
TODO: this fails with numbered sections.

Re-enable "Wiki-Links":

Hyperlink references without target default to pages of the local wiki. 
"Forward-port" from moin 1.9.; adapt to the new program structure: 
The `Writer` class is no longer used, so we need a custom parser to define and register a "resolver" method for hyperlink references without a matching target.

Call `docutils.core.publish_doctree` with the custom parser, specify a "source_path" identifier (for improved error messages).

TODO:
Check, whether the `rst_in.Writer` class can be removed. 
Consider a "docutils_settings" variable for configuration from conf.py.


RogerHaase:
@gmilde: https://moin-20.readthedocs.io/en/latest/devel/development.html#install-pre-commit-hooks

Then, after staging your changes: https://moin-20.readthedocs.io/en/latest/devel/development.html#review-your-working-solution

gmilde:
Thank you for the feedback. I will prepare another PR.
I did not have success with `./m tests`, though (`.m: Kommando nicht gefunden.`)
Managed with `pytest .` instead.

UlrichB22:
> Thank you for the feedback. I will prepare another PR.
You can just commit your updates to the same branch and push it to update this PR.

> I did not have success with `./m tests`, though (`.m: Kommando nicht gefunden.`)
There is a typo in your error message: it should be ./m and you need to run it in the Git root directory. The script named “m” is created the first time quickinstall.py is called.

gmilde:
> You can just commit your updates to the same branch and push it to update this PR.

I'd rather start new with separate PRs for "wiki links" and "internal links".

gmilde:
> You can just commit your updates to the same branch and push it to update this PR.

I'd rather start new, with separate PRs for "wiki links" and "internal links".

Issue #1911, opened 2025-05-07
Focus theme as the new default theme


cintek:
One way to convince people that MoinMoin is good software and better than other wiki software is good usability. To improve it for future moin2 users we recommend to set the focus theme as the new default theme. Often when systems like this wiki are set up the default settings are not changed so moin2 should offer the theme with the best usability right away per default.

cintek:
Thanks that you resolved some of the issues already.

> 2\. on a small screen: the actions are in one horizontal bar und need scrolling (screenshots below).
>         Is this intended?

For the first version it was okay but I think it should be changed and the smaller view needs some more love in general. The area at the top takes too much space.

I will take a look at this and the other items.

gmilde:
The styling of tables without border lines may be elegant for simple tables but makes it hard to comprehend complex tables.
See, for example the rST help page (in a local wiki it's http://localhost:5000/help-en/rst#Tables).

Moin supports the `no-borders` "userstyle CSS class, so it is (relatively) easy to disable table borders, but it requires custom CSS to enable table borders.

roland-ruedenauer:
Moin supports adding alias names to wiki items. The rendering of items with alias names in the pagetrail using the Focus theme looks somehow unconventional:

<img width="787" height="676" alt="Image" src="https://github.com/user-attachments/assets/062c4c1e-5520-4e9a-87a0-382f64d61c59" />

Other themes display the alias names in a popup window when hovering on the item name.

cintek:
> The styling of tables without border lines may be elegant for simple tables but makes it hard to comprehend complex tables.

Indeed, we discussed about this already internally and we also think that there should be something to make is easier to distinguish between different lines. This could also be achieved with alternating background color.

> Moin supports adding alias names to wiki items. The rendering of items with alias names in the pagetrail using the Focus theme looks somehow unconventional:

Thanks for the hint.

gmilde:
> [...] there should be something to make is easier to distinguish between different lines. This could also be achieved with alternating background color.

Alternating background colour is less suited for tables with rowspans, though. 
This is why I recommend a fine grid as the default and use of the pre-defined "userstyle" CSS classes "no-borders" and "zebra" for alternate table styles.

roland-ruedenauer:
Focus theme displays the page content (the div element with id "moin-content") centered in the screen area to the right of the sidebar panel.
[...]

PR #1922, opened 2025-05-21
Focus: fix itemview icons, rm globe-rotate-left-solid, add login icon

UlrichB22:
... 
This change also moves the icon mapping from themes get_endpoint_iconmap() to constants/misc.

The '-o' suffix for Fontawesome icons like "fa-trash-o" has been removed, this seem to be a leftover from a version prior 6.x.
This also fixes missing icons in itemviews for different other themes.

Related to #1911 no. 3 and no. 7


UlrichB22:
Also related to #1923.

roland-ruedenauer:
> Regarding the use of Font Awesome icons:
Making use of the regular icons instead of the solid ones would probably make the Basic theme look a bit nicer.

As noted in the PR fixing the Basic theme issues, the regular icons would match the previously used "-o" ones. Imo they also look better. I suggest having a look and judge by yourself.

Issue #1912, opened 2025-05-07
Redundant code in themes


UlrichB22:
Templates in the subdirectories of themes contain redundant code that may be outdated. 

E.g., if you run a diff between templates/show.html and themes/focus/templates/show.html, you will find 
 * some changes since October 2023 that are missing in the focus theme
 * some changes related to the new theme.

I have not checked the other template files.

It would be desirable if the differences could be controlled by a variable. This should be possible as the focus theme is now part of the default codebase.

RogerHaase:
@UlrichB22 is there anything left to do here?

Issue #297, opened 2012-11-27
HTML source output is ugly


ThomasWaldmann:
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/297) by RogerHaase (Bitbucket: [RogerHaase](https://bitbucket.org/RogerHaase), GitHub: [RogerHaase](https://github.com/RogerHaase)).**

----------------------------------------

The generated HTML source has too much whitespace, too few line endings, and not enough whitespace in the right places.

Is there a place where a post processor could intercept output and format it properly?

PR #1941, opened 2025-06-05
Theme consolidation regarding template show.html

UlrichB22:
- Remove show.html template from theme 'focus', related to #1912.
- Reduce consecutive line breaks and beautify html in itemviews template, related to #297.

This change removes a redundant template, but adds an additional call to theme() for themes other than 'focus'.
I would be happy to receive your feedback.

### DIFF
--- a/src/moin/templates/itemviews.html
+++ b/src/moin/templates/itemviews.html
@@ -7,13 +7,12 @@
     Comment and Transclusion links may be hidden by Javascript if item lacks content.
 #}
 
-{% set item = item %}
-{% set not_trash =  not item_is_deleted if item_is_deleted is defined else True %}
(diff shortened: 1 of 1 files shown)


RogerHaase:
Changes look OK, but with window width  less than 930 pixels the navigation bar is hidden and replaced with a "toggle menu" hamburger in upper right. Clicking it has no effect.

UlrichB22:
I am getting this "toggle menu" as well with theme focus and window width less than 930 pixels. Clicking the button opens the navigation area as expected. Looks like problems with Javascript in your environment. Can you check the same with branch master? It should have the same behavior.

UlrichB22:
I will close this PR because of conflicts and create a new one for removal of show.html from theme 'focus'.

PR #1959, opened 2025-07-13
Make use of meta data key constants

roland-ruedenauer:


src/moin/items/__init__.py
```diff
@@ -1568,8 +1570,8 @@ def do_modify(self, item_may=None):
                 item = Item.create(template_name)
                 form = self.ModifyForm.from_item(item)
                 # replace template name with new item name and remove TEMPLATE tag
-                form["meta_form"]["name"] = self.name
-                form["meta_form"]["tags"].remove(TEMPLATE)
+                form["meta_form"][NAME] = self.name
+                form["meta_form"][TAGS].remove(TEMPLATE)
```
ThomasWaldmann, line 1574: These are keys into a **form** dict, not into a revision meta dict.

So, while it is the same value, it's a different use case.

src/moin/items/__init__.py
```diff
@@ -1651,7 +1653,7 @@ def do_modify(self, item_may=None):
                                 )
 
                     # save the new revision, unlock, delete draft
-                    contenttype_qs = request.values.get("contenttype")
+                    contenttype_qs = request.values.get(CONTENTTYPE)
```
ThomasWaldmann, line 1656: Similar here: stuff from URL, not from revision meta.

src/moin/cli/maint/modify_item.py
```diff
@@ -22,7 +22,19 @@
...
+    ITEMID,
+    DATAID,
+    NAMESPACE,
+    REVID,
+    SIZE,
+    PARENTID,
+    REV_NUMBER,
+    MTIME,
+    NAME,
+)
```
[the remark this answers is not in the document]
roland-ruedenauer: This is the result of auto-formatting by black.

src/moin/items/__init__.py
roland-ruedenauer: Agree. Reverted.

PR #1964, opened 2025-07-25
Config: add a dict-like getter

sebix:
makes the access to optional configuration options much easier

### DIFF
--- a/src/moin/config/default.py
+++ b/src/moin/config/default.py
@@ -266,6 +266,15 @@ file. It should match the actual charset of the configuration file.
         """Make it possible to access a config object like a dict"""
         return getattr(self, item)
 
+    def get(self, item, default=None):
+        """Dict-like get method.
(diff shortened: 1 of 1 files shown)


src/moin/config/default.py
```diff
@@ -266,6 +266,15 @@ def __getitem__(self, item):
...
 
+    def get(self, item, default=None):
+        """Dict-like get method.
+        Return the value for key if key is in the dictionary, else default.
+        If default is not given, it defaults to None, so that this method never raises a KeyError."""
+        try:
+            return getattr(self, item)
+        except AttributeError:
+            return default
+
```
ThomasWaldmann, line 277: IIRC the idea of the DefaultConfig is that it has defaults for all supported attributes, so one would never get this `AttributeError`.

Having it in this class is also useful because it is one place where settings and their default can be looked up.

PR #1996, opened 2025-09-28
babel: move config to pyproject.toml

ThomasWaldmann:
Babel >= 2.16.0 (from 2024-08) supports this.

https://github.com/python-babel/babel/pull/1108


UlrichB22:
Before running `pybabel extract ...` you need to update Babel to a version >= 2.16.0.

PR #1999, opened 2025-09-30
Some Itemtype related code changes

roland-ruedenauer:
Userprofile items: set properties ITEMTYPE and REV_NUMBER at creation.
Use existing itemtype constants in blog.py.

### DIFF
--- a/src/moin/_tests/test_user.py
+++ b/src/moin/_tests/test_user.py
@@ -11,8 +11,9 @@
 from flask import g as flaskg
 
 from moin import user
+from moin.constants.itemtypes import ITEMTYPE_USERPROFILE
 from moin.items import Item
-from moin.constants.keys import ITEMID, NAME, NAMEPREFIX, NAMERE, NAMESPACE, TAGS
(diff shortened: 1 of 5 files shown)


UlrichB22:
Existing users still do not have ITEMTYPE and REV_NUMBER in there meta data. What is the background for this change?

roland-ruedenauer:
The motivation here is improving consistency in the foundations of the data model.

* Why set ITEMTYPE for most of the entries but not Userprofile?
* Searching for `meta[REV_NUMBER]` indicates pieces of existing code expect the property to exist

Especially the REV_NUMBER looks confusing to me. Sometimes it is missing for an item, while for newer revisions it might me present. Then it is unclear whether it is 0 based or not.

UlrichB22:
Should we change the `'moin maint-validate-metadata'` command to set ITEMTYPE and REV_NUMBER for existing wikis?  

https://github.com/moinwiki/moin/blob/464244c15bcecbea9a61f33a85593ac15cac6d02/src/moin/cli/maint/modify_item.py#L289-L292

roland-ruedenauer:
> Should we change the `'moin maint-validate-metadata'` command to set ITEMTYPE and REV_NUMBER for existing wikis?

In my opinion: yes, for sure

You may also check the content import functionality from old wiki instances does "the right thing" too.

PR #2004, opened 2025-10-07
Pass variable 'subitem_target' instead of 'item_name' into search templates.

roland-ruedenauer:
Variable 'item_name' is already used for other purposes in layout.html.

Note that 'item_name' might be undefined when navigating to a non item-related view like search. This required an additional change in file layout.html of the basic theme.


roland-ruedenauer:
Basic search functionality worked for me. Not sure about the linking for blog item search is correct. I've actually never used this functionality. Can someone check this, please?

PR #2008, opened 2025-10-18
Correct template rendering

roland-ruedenauer:
No longer modify the Jinja environment globals with each request processing. Doing otherwise caused issues with objects not being released as expected (e.g. "clock") because they were still being referenced somewhere in cached content created by the rendering engine.

> Environment globals should not be changed after loading any templates, and template globals should not be changed at any time after loading the template. Changing globals after loading a template will result in unexpected behavior as they may be shared between the environment and other templates.

See <https://jinja.palletsprojects.com/en/stable/api/#the-global-namespace>

A context procesor is used instead to inject common variables to the rendering engine. Template import statements must specify "with context" when accessing context variables like e.g. "gen".

### DIFF
--- a/src/moin/app.py
+++ b/src/moin/app.py
@@ -29,13 +29,15 @@ from flask_theme import setup_themes
 from jinja2 import ChoiceLoader, FileSystemLoader
 from whoosh.index import EmptyIndexError
 
-from moin.utils import monkeypatch  # noqa
-from moin.utils.clock import Clock
 from moin import auth, user, config
 from moin.constants.misc import ANON
 from moin.i18n import i18n_init
-from moin.themes import setup_jinja_env, themed_error
+from moin.search import SearchForm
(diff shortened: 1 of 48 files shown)


roland-ruedenauer:
To observe the problematic behaviour with the existing code you could modify function `teardown_wiki` at https://github.com/moinwiki/moin/blob/master/src/moin/app.py#L389 and inspect the log output when rendering a wiki page.

```
import gc

...

def teardown_wiki(response):
   ...
        clock = flaskg.pop("clock", None)
        if clock is not None:
            clock.stop("total", comment=request_path)
            gc.collect()
            referrers = gc.get_referrers(flaskg.clock)
            rc = sys.getrefcount(flaskg.clock)
            logging.debug(f"obj: rc is { rc }")
            logging.debug(f"obj: referrers are { referrers }")
            del clock
   ...
```

PR #2022, opened 2025-11-02
rST documentation update

gmilde:
Start updating/correcting/extending the documentation for Moin reStructuredText support. 
There is still much work to do (with both fixing problems and documentation)

Please have a look, if style and terminology are consistent with Moin help. 
Suggestions and advise is welcome.

I suggest using either the full name "reStructuredText" or the abbreviation "rST":

* a web search for  "REST", "ReST", "reST" (and even for "reST markup") will show "[Representational State Transfer](https://sv.wikipedia.org/wiki/Representational_State_Transfer)" as first hit.
* search results for "rST" are not much better, but it is more similar to the usual file extension and shorter. Searching for "rst markup" brings enlightenment :)

### DIFF
--- a/src/moin/converters/rst_in.py
+++ b/src/moin/converters/rst_in.py
@@ -127,10 +127,8 @@ class NodeVisitor:
         # use "attention" for generic admonitions, cf.
         # http://docutils.sourceforge.net/docs/ref/rst/directives.html#generic-admonition
         self.open_moin_page_node(moin_page.admonition({moin_page.type: typ}), node)
-        self.header_size += 1
 
     def depart_admonition(self, node=None):
-        self.header_size -= 1
         self.close_moin_page_node()
 
     # see http://docutils.sourceforge.net/docs/ref/rst/directives.html#specific-admonitions
@@ -643,6 +641,8 @@ class NodeVisitor:
         self.header_size -= 1
 
     def visit_sidebar(self, node):
+        # TODO: render sidebar "set off from the rest of the document somehow, typically with a border."
+        # Sidebars typically “float” to the side of the page.
         pass
 
     def depart_sidebar(self, node):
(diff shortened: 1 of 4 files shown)


gmilde:
The failing test stumbles about an
> ERROR scrapy.downloadermiddlewares.retry:117 Gave up retrying <GET http://localhost:8080/robots.txt> 

I don't know how to solve this :(

roland-ruedenauer:
```
>       assert len(failures) == 0
E       assert 1 == 0
E        +  where 1 = len([CrawlResult(url="http://127.0.0.1:9080/Example%20Page", from_url="http://127.0.0.1:9080/help-en/rst", from_text='Exam...sponse_exc="<twisted.python.failure.Failure scrapy.spidermiddlewares.httperror.HttpError: Ignoring non-200 response>")])
```

Error code 404. Meaning you added a link to a non existing help page in your updated RST documentation. Search for "Example Page". Did you forget to add a new file?

gmilde:
Actually the link to a non-existing wiki item was intended: 
The link is part of the demonstration of the wiki-internal link syntax:
a link to an item in the default namespace with a space in its name.
It is the expected outcome that clicking on the link will open the "create new item" dialogue. I did not know that there is a test for link target existence (I agree that this makes sense in the "normal" case).

Is it safe to assume the "wiki.local:Home" item is available? (It is missing in my minimal "dummy wiki" created for updating the help metadata files.)

roland-ruedenauer:
The invocation of `moin create-instance --full` will create the page `Home` in LoadWelcome(). Without option `--full` the wiki instance will be created totally empty.

See https://github.com/moinwiki/moin/blob/master/src/moin/cli/maint/modify_item.py#L353

gmilde:
It seems the CI tests invoce `moin create-instance --full`...

> All checks have passed

PR #2010, opened 2025-10-21
Feat : Added dark theme as per system settings [#1502]

omjamnekar:
# Add Automatic Dark Mode Based on System Preference [#1502]

This update introduces **automatic dark mode** that responds to users’ browser or operating system settings via the CSS media query `@media (prefers-color-scheme: dark)`.

---

##  Changes Implemented

- Added a **dark color palette** (80+ CSS variables) in `variables.css`  
- Updated `common.css` and `theme.css` to style all UI elements for dark mode  
- The interface now **automatically switches** between light and dark themes based on the system/browser setting  
- Removed previous **manual theme toggle button** and related script (`theme-toggle.js`)  
- Simplified logic to rely purely on **native browser behavior** — no user interaction or `localStorage` needed  

---

##  Benefits

- [x] **automatic dark mode** (no configuration required)   
- [x] **Power savings** on AMOLED displays  
- [x] **Consistent experience** across desktop and mobile browsers  
## Night time

<h2>Desktop</h2>
Screenshot 2025-10-21 013722
Screenshot 2025-10-21 013750
<h2>Day time</h2>

Screenshot 2025-10-21 100626
Screenshot 2025-10-21 101347

Issue #1502, opened 2023-09-07
Automatic light/dark theme switching with `prefers-color-scheme` CSS media query


nekohayo:
It would be nice if MoinMoin's default stylesheet / theming system would automatically offer a dark stylesheet at nighttime. GitHub itself does this, for example.

My operating system does the switch for me everyday at the correct times, and my web browser follows suit, telling webpages' CSS mediaqueries that dark is now "preferred" or not.

MoinMoin could either automatically offer a dark stylesheet at nighttime using this trick, i.e. "automatic" light and dark themes "based on browser/OS setting", as the default behavior (rather than needing users to turn this on as a setting), though maybe some users will want the ability to override the automatic behavior.

You can detect the system's state using this simple piece of CSS, such as:

`@media (prefers-color-scheme: dark) { blah blah }`

This might be interesting not only for desktop computers, but also for people using the web interface on mobile devices; that way if they're reading messages/directions at night on the street or in a dimly lit location, they won't get blinded by the white screen. It also tends to consume less power on AMOLED displays.

omjamnekar:
@UlrichB22  i would like to work on this

UlrichB22:
@omjamnekar, thanks for your support. I like the new dark theme.

Can you check the other themes? In the “Focus” theme e.g., the input fields are now black, but everything else seems unaffected by the changes. This is a problem with the use of variable names in the different designs. Some variables are used globally. Do you have any ideas for a naming concept that includes global variables and theme-dependent variables?

----

<img width="794" height="539" alt="Image" src="https://github.com/user-attachments/assets/6a75b463-6881-494c-b715-0ce54ecb3380" />

UlrichB22:
@omjamnekar are you still working on this issue?

The PR still needs some optimization, such as 
- using the newly introduced variable names in the common.css file and 
- fixing side effects on themes other than the default “topside” theme.

omjamnekar:
I might not be able to work on it

UlrichB22:
> I might not be able to work on it

Thanks for supporting moin. I'll try to finish the feature.

srikaaviya:
Shall I work on this?

UlrichB22:
@srikaaviya, yes, you're welcome. Please share your ideas with us before you start coding.

The basis is the “topside” theme. I've noticed that the code highlighting features are ugly and buggy in dark mode. You can test with `'help-en/OtherTextItems'` and switch to dark mode in the developer mode of your browser.

[...]

PR #2026, opened 2025-11-04
Feature: Add dark theme as per system settings

UlrichB22:
Related to #1502.
Replaces #2010.

Moved all code from variables.css and common.css to theme/topside/theme.css to avoid side effects on other themes like theme 'focus'. This can be consolidated if every theme has a dark mode.


UlrichB22:
This PR is based on the work of @omjamnekar. Many Thanks for your support.

roland-ruedenauer:
Wouldn't it make sense to replace color codes with variable references as you've done for the `body` styling?

UlrichB22:
> Wouldn't it make sense to replace color codes with variable references as you've done for the `body` styling?

You're right, but that requires some more work. There are no predefined variables for the other colors. This is just a minimal version to use the existing PR. 

The next step will be to add this feature to the other themes and use the same color set with unique names.

PR #2037, opened 2025-11-08
Improve existing test code

roland-ruedenauer:
* remove moin.i18n.force_locale; use force_locale provided by flask_babel; make existing test case run by default
* replace setup_class methods used in test_markdown_in(_out) with pytest fixtures performing orderly cleanup
* correct tests in TestProtectedIndexingMiddleware (unprotected middleware was used instead of the protected one)
* get rid off autouse fixture app; use @pytest.mark.usefixtures annotation in test cases to make the required context explicit; this change also avoids having to instantiate an application context for every unit test execution
* avoid autouse class method fixtures; use mark.usefixtures (possibly with more than one argument)
* change names of test methods along the way to be conformant to the usual naming conventions
* provide utils in moin.apps._tests.utils to be used with new pytest fixture client for writing tests using the Flask test client
* fixtures returning None are prefixed with an underscore
* get rid of "_dummy" test methods in TestProtectingMiddleware by introducing base class TestIndexingMiddlewareBase

### DIFF
--- a/src/moin/_tests/test_forms.py
+++ b/src/moin/_tests/test_forms.py
@@ -5,6 +5,8 @@
     MoinMoin - moin.forms tests.
 """
 
+import pytest
+
 import datetime
 import json
 from calendar import timegm
@@ -51,6 +53,7 @@ def test_datetimeunix():
     assert d.u == ""
 
 
+@pytest.mark.usefixtures("_req_ctx")
 def test_validjson():
     """
     Tests for changes to metadata when modifying an item.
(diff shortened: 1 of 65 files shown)


roland-ruedenauer:
Sorry, this one got a bit large und things might be mixed up. I worked on this branch for quite a while and it got what it is now.

Anyways, one positive side effect is: the execution time of "./m tests" is brought down considerably. At least for me when testing locally under Ubuntu 24.04 (WSL) it went down by ~40%.

RogerHaase:
Test run time dropped ~30% on Windows 10

Issue #2043, opened 2025-11-10
@import "notice"; and @import "bootstrap"; fails when running `./m css`


RogerHaase:
Documenting need to fixup missing imports after https://github.com/moinwiki/moin/pull/1908:

Convert Basic theme from Boostrap 3 to Boostrap 4 and use sass instead of lessc (also fixes #1770) #1908

If you get the @import "notice"; and @import "bootstrap"; errors running `./m css` then do:

1. point your browser to https://github.com/twbs/bootstrap/tree/v4.5.3/scss (optional)
2. modify the browser url to https://ssgithub.com/twbs/bootstrap/tree/v4.5.3/scss
3. click the download button to save a zip file to a convenient location
4. unzip the file to xstatic/pkg/bootstrap/data creating a scss directory as a sibling to the existing css and js directories
5. rerun `./m css`

roland-ruedenauer:
In #1908 I mentioned:

> A new release of XStatic-Bootstrap should provide the required scss files.

Unfortunately the process of releasing updated XStatic packages got stuck and the boostrap SCSS files are still missing now months later.

My suggestion made in #1895 would still be an (imo good) option to solve this obstacle.

RogerHaase:
docs updated https://moin-20.readthedocs.io/en/latest/devel/development.html#add-more-tools-exercise-tools

PR #2059, opened 2025-11-25
UI fixes related to flash/validation messages when saving user settings 

roland-ruedenauer:
Fixes two issues we currently have with saving user settings
* validation error message getting lost when saving user settings (caused by a client reload call made even if errors were present)
* avoid displaying the same flash messages twice (XHR case: the client code is fully responsible dealing with flash messages)

### DIFF
--- a/src/moin/static/js/common.js
+++ b/src/moin/static/js/common.js
@@ -469,6 +469,8 @@ MoinMoin.prototype.enhanceUserSettings = function () {
         buttonDotAnimation = setInterval(buttonRunAnimation, 500);
         buttonRunAnimation();
 
+        MoinMoin.prototype.clearFlashMessages();
+
         // send the form to the server
         $.post(form.attr('action'), form.serialize(), function (data) {
             clearInterval(buttonDotAnimation);


UlrichB22:
I checked the Usersettings view and the User CSS URL in the tab Appearence. When entering "@wrong.css",  
it now shows the applied value in red and the error message. 

The next problem is the size and location of the message, which is not optimal.

image

roland-ruedenauer:
Thanks for testing this.

The latest commit tries to improve upon what you've shown above. Please note that I didn't check interference with other themes. Further UI improvements should probably handled in a separate merge request.

image

UlrichB22:
Test results for different themes (User settings > Appearance > User CSS URL field):

- Topside theme: now looks good
- Focus theme: input field and error message are good, but there are two flash messages: 1. Your changes have been saved. 2. Nothing saved.
- Modernized theme: same as Focus theme
- Basic theme: the input field changes the padding and radius and looks ugly, the flash message at the top is correct.

----

**Modernized theme**

image

----

**Basic theme**

image

roland-ruedenauer:
So basic theme styling for moin-error class needs to be adjusted as well.

The two flash messages look really weird. I'll have look.

roland-ruedenauer:
@UlrichB22 Please have another look

UlrichB22:
@roland-ruedenauer thanks for fixing this issue. My test with CSS_URL and the password change were successful for all standard themes. 

For theme basic the CSS files need to be regenerated with `./m css` and additional instructions in #2043.

roland-ruedenauer:
Hmm ... I've updated the CSS files as denoted in #2043, which btw might be closed now that the documentation was updated (@RogerHaase ).

PR #2062, opened 2025-12-01
Replace user option CSS_URL with CSS_FILE for security reasons

### LINKED ISSUE #1956: User CSS URL and wikiconfig option for custom CSS
In the user settings there is an option to specify a “User CSS URL”. This can lead to a CSS injection and should IMO be disallowed. 
We can either
 * remove this option or
 * change it to a local file or path relative to wiki_local. However, this only allows the use of CSS files provided by the webadmin/wikiadmin.
 
Also, I would like to have a wikiconfig option to specify a custom CSS file or path relative to wiki_local. This makes custom styles easier as it is not necessary to copy and adapt the snippets.html for simple changes like logos, fonts and colors. This file would need to be maintained if there are changes to the snippets.html source.

UlrichB22:
Fixes #1956.

Many thanks to @roland-ruedenauer for his help with the validator.

### DIFF
--- a/docs/user/accounts.rst
+++ b/docs/user/accounts.rst
@@ -129,7 +129,7 @@ Theme name
 CSS file
  If you want to style MoinMoin with custom Cascading Style Sheets (CSS), enter a filename for your
  custom stylesheet here. Custom CSS provides an advanced level of control over appearance of
- MoinMoin pages. The file needs to be in the “wiki_local” subdirectory of the wiki instance.
+ MoinMoin pages. The file needs to be in the wiki_local subdirectory of the wiki instance.
  Ask your wiki administrator for assistance.
 
 Number of rows in edit textarea
--- a/src/moin/apps/frontend/views.py
+++ b/src/moin/apps/frontend/views.py
@@ -2481,9 +2481,9 @@ class FileNameValidator(Validator):
         filename = element[self.element_name].value
         base_file_name = os.path.basename(filename)
         if (
-            not filename == base_file_name
-            or len(str(base_file_name)) > 255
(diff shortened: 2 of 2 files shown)


docs/user/accounts.rst
```diff
@@ -126,10 +126,11 @@ Appearance settings allow you to customize the look and feel of the wiki.
 Theme name
  The bundled MoinMoin wiki theme which you would like to use.
 
-User CSS URL
- If you want to style MoinMoin with custom Cascading Style Sheets (CSS), enter a URL for your
+CSS file
+ If you want to style MoinMoin with custom Cascading Style Sheets (CSS), enter a filename for your
  custom stylesheet here. Custom CSS provides an advanced level of control over appearance of
- MoinMoin pages.
+ MoinMoin pages. The file needs to be in the “wiki_local” subdirectory of the wiki instance.
```
roland-ruedenauer: > “wiki_local”

Out of interest: the use of opening/closing double quotes here is probably legit, but was it intended?

src/moin/apps/frontend/views.py
```diff
@@ -2467,6 +2466,29 @@ def validate(self, element, state):
...
+
+    def validate(self, element, state):
+        if not self.element_name or self.element_name not in element:
+            raise RuntimeError("form element_name not configured or not found.")
+        if not element[self.element_name].valid:
+            return False
+        filename = element[self.element_name].value
+        base_file_name = os.path.basename(filename)
+        if (
+            not filename == base_file_name
```
roland-ruedenauer: Suggest using "!=" here.

PR #2077, opened 2025-12-17
Sendmail: add timeout and SMTP_SSL support, remove mail_sendmail

### LINKED ISSUE #2073: Sendmail module needs to be revised.
The sendmail module needs to be revised.
 
These are the open topics:
 
1. The name of the website is not replaced correctly in the subject line of the email. 
2. Some email providers require smtp_ssl for the connection, which has not yet been implemented.
3. The timeout parameter for the smtplib.SMTP class should be used to avoid endless hanging connections.
4. The /usr/sbin/sendmail function has not yet been implemented, and moin outputs a traceback when the mail_sendmail configuration setting is used.

I recommend removing the mail_sendmail configuration option. Please advise.

This is a preparation for #2056.

UlrichB22:
Included changes are

- Fix the sitename substitution in the subject of the password recovery mail
- Add timeout to smtplib.SMPT call
- Remove incomplete mail_sendmail cli feature
- Use SMTP_SSL if the SMTP port is 465

Fixes #2073.

I tested the UI password recovery feature with a server with port 465 and some invalid configs.
Please help with tests for other SMTP server and the subscription notification feature. Thanks.

### DIFF
--- a/src/moin/mail/sendmail.py
+++ b/src/moin/mail/sendmail.py
@@ -85,15 +85,19 @@ def sendmail(subject, text, to=None, cc=None, bcc=None, mail_from=None, html=Non
     msg["Auto-Submitted"] = "auto-generated"  # RFC 3834 section 5
 
     # Connect to SMTP server
-    host, port = (cfg.mail_smarthost + ":25").split(":")[:2]
-    port = int(port)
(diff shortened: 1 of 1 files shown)


UlrichB22:
The subscriber notification feature has been successfully tested. Two different users were required for testing.

Issue #2065, opened 2025-12-08
Upload of name.xml.xz file denied as subitem


sebix:
1. Go to any page
2. Go to the subitems
3. Click New Item
4. Click Browse
5. Select an xz-compressed XML file named `name.xml.gz`. The `.xml.gz` part seems to be crucial.
6. The response is: `UnicodeDecodeError, upload failed, not a text file, nothing saved: 'short.xml.xz'. Try changing the name.`

The wiki allows uploading binary files, so an xz file should be allowed. And xz is clearly not a text file, so the error message is wrong or misleading.

Uploading the file as `short-xml.xz` works, so there may be XML parsing triggers where they shouldn't be.

roland-ruedenauer:
Thanks for dropping a note and your clarifications.

I've provided a fix in #2078. You may check again, if the issue is resolved with the contained changes.

PR #2078, opened 2025-12-17
Fix issue #2065

roland-ruedenauer:
The return value _encoding_ was ignored when using [mimetypes.guess_type(url)](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type) to determine the mime-type from the supplied filename.

### DIFF
--- a/src/moin/utils/_tests/test_mimetype.py
+++ b/src/moin/utils/_tests/test_mimetype.py
@@ -48,6 +48,14 @@ class TestMimeType:
             (".jpeg", "image/jpeg"),
             (".png", "image/png"),
             (".svg", "image/svg+xml"),
+            (".tar.bz2", "application/x-tar"),
+            (".tar.gz", "application/x-tar"),
+            (".tar.xz", "application/x-tar"),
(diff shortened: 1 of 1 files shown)


src/moin/utils/mimetype.py
```diff
@@ -79,6 +79,9 @@
     ("application", "x-tex"): ("text", "tex"),
     ("application", "javascript"): ("text", "javascript"),
     ("application", "x-dos-batch"): ("text", "bat"),
+    # Non-standard MIME types used on some platforms
+    ("application", "x-gzip"): ("application", "gzip"),  # Windows, macOS
+    ("application", "x-zip-compressed"): ("application", "zip"),  # Windows
```
sebix: I'm not sure that is correct.

x-zip is normally abbreviated as xz and is not equal to zip

roland-ruedenauer:
Well, I should have spent some time on adding tests. Pushed the next attempt to fix the issue.

sebix (approved):
Thanks! I can confirm #2065 is fixed with the PR.

PR #2082, opened 2025-12-22
Fix CSS issues

roland-ruedenauer:
Mainly fixes issues with the styling of the Basic theme. See the commit comments for details.

@gmilde The rST styling should still be correct. Can you have a look, please?

### DIFF
--- a/src/moin/cli/maint/dump_html.py
+++ b/src/moin/cli/maint/dump_html.py
@@ -261,7 +261,7 @@ def Dump(directory="HTML", theme="topside_cms", exclude_ns="userprofiles", user=
             # work around differences in basic and modernized theme layout
             # TODO: this is likely to break as new themes are added
             if theme == "basic":
-                start = '<div class="moin-content" role="main">'  # basic
+                start = '<div class="moin-content flex-row" role="main">'  # basic
                 end = '<footer class="navbar moin-footer">'
                 div_end = "</div>"
             else:
--- a/src/moin/themes/basic/templates/layout.html
+++ b/src/moin/themes/basic/templates/layout.html
@@ -184,6 +184,7 @@
                 {{ after_header }}
                 {{ utils.breadcrumbs(theme_supp.path_breadcrumbs, title_name) }}
 
+                {# Note: dump_html.py might require adaptation when modifying the following line #}
                 <div class="moin-content flex-row" role="main">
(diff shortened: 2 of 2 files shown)


gmilde:
I suggest a different approach for the "sidebar" fix (first commit):
Let's use different classes for a "moin-sidebar" and a "moin-rst-sidebar" -- as these are actually very different objects. There is already a precedence for "rst object emulation rules" like `moin-rst-fieldlist` etc.

In any case, we don't need additional classes for the "moin-content", as a CSS rule may also use an element's ID as selector.

roland-ruedenauer:
The class "moin-content" wasn't added by me. It was already part of the Basic theme.

Adding a class "main-rst-sidebar" of course is a viable solution too. Initially I was tempted to to the same, but then I thought it might not be the best idea to introduce content format specific CSS classes.

gmilde:
> Adding a class "moin-rst-sidebar" of course is a viable solution too. Initially I was tempted to to the same, but then I thought it might not be the best idea to introduce content format specific CSS classes.

I would describe the "moin-rst-*" classes as emulating "elements known from reStructuredText and documented in [The Docutils Document Tree](https://docutils.sourceforge.io/docs/ref/doctree.html#contents)".

Currently we have "moin-rst-fieldlist", "moin-rst-optionlist", and
"moin-rst-attribution", all by  R. Haase, 2016.
Like "admonitions", they are not solely for use with content marked up with rST.

Considering the use in the "div.moin-content" part of the page,
we may consider removing the prefix "moin-" (leaving "rst-fieldlist" etc.)
and leaving the "moin-" namespace for Moin-added header, footer, and sidebar.

gmilde:
PR #2085 implements an alternative to the addition of the class "moin-content" in 4 layout templates. (commit 1).

The `.scss` update is needed in any case.

RogerHaase:
@roland-ruedenauer do you agree with @gmilde to close this pr in favor of #2085?

gmilde:
Mind

> @roland-ruedenauer do you agree with @gmilde to close this pr in favor of #2085?

Mind, that commits 2 and 3 are still required.

roland-ruedenauer:
I've removed the sidebar CSS fix from the MR although to me the combined use of `"html-aside rst-sidebar"` in #2085 doesn't look like an improvement. And now we have `moin-rst-*` and `rst-*` CSS classes.

Don't forget to regenerate `theme.css` after updating the Basic theme.

PR #2083, opened 2025-12-26
Small fixes for HTML-in.

gmilde:
Remove "strong" from `simple_tags`. It is already in `symmetric_tags`.

Remove "var" from `ignored_tags`. It's a bad idea to ignore the content of a `<var>` element (the name or symbol of a variable or constant).

Fix docstring for `visit_xhtml_inline()`:
the tag name of `inline_tags` is no longer stored in the "element" attribute but as "class" value following the convention `"html-" + tag.name` (cf. the HTML converter tests).

Use an "f-string" to form the "class" value in `visit_xhthml_inline()`.

Add some comments.

### DIFF
--- a/docs/admin/configure.rst
+++ b/docs/admin/configure.rst
@@ -1476,9 +1476,9 @@ Mail configuration
 
 Sending Email
 --------------
-Moin can optionally send email. Possible uses:
+Moin can optionally send emails. Possible uses:
 
-* send out item change notifications
+* send notifications about item changes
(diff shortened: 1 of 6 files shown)


gmilde:
The second commit goes a bit further:

Add tags with Moin-equivalents to the
`symmetric_tags` (homonymous Moin elements) or
`simple_tags` (differently named Moin elements).
Remove the specific methods that used to handle them.

Fix some typos in comments and docstrings.

PR #2085, opened 2025-12-30
rST in: fix side-effect of rST "sidebar" styling.

gmilde:
Class "moin-sidebar" is used for the sidebar panel in Basic theme.

Using it also for rST sidebar elements (in commit b67464a9) interferes with the Basic theme's sidebar panel :(

Use class value "html-aside" for elements with "content outside the flow of the main content of the document" (like an HTML `<aside>` element).

Add class value "rst-sidebar" to the element generated by a reStructuredText "sidebar" directive.

TODO: drop CSS rule for rST sidebars and leave styling to custom CSS?

This commit is an alternative to commit 1 in PR #2082.

### DIFF
--- a/src/moin/converters/_tests/test_rst_in.py
+++ b/src/moin/converters/_tests/test_rst_in.py
@@ -509,11 +509,11 @@ text""",
         # Topics, Sidebars, and Rubrics
         (
             ".. topic:: Topic Title\n   :class: custom\n\n   topic content",
-            '<div xhtml:class="moin-aside custom"><p xhtml:class="moin-title">Topic Title</p><p>topic content</p></div>',
+            '<div xhtml:class="html-aside custom"><p xhtml:class="moin-title">Topic Title</p><p>topic content</p></div>',
         ),
         (
             ".. sidebar:: Sidebar Title\n   :subtitle: Sidebar Subtitle\n   :class: float-right\n\n   sidebar content",
-            '<div xhtml:class="moin-aside moin-sidebar float-right">'
+            '<div xhtml:class="html-aside rst-sidebar float-right">'
             '<p xhtml:class="moin-title">Sidebar Title</p>'
             '<p xhtml:class="moin-subheading">Sidebar Subtitle</p>'
             "<p>sidebar content</p></div>",


gmilde:
In https://github.com/moinwiki/moin/pull/2082#issuecomment-3710696251,
Roland Rüdenauer wrote:
> to me the combined use of "html-aside rst-sidebar" in #2085 doesn't look like an improvement.  And now we have moin-rst-* and rst-* CSS classes.

The class values `html-aside` and `rst-sidebar` were selected to replace the current values `moin-aside` and  `moin-sidebar` with the following rationale:

**`moin-aside` -> `html-aside`**
The first class value indicates that the `<moin:div>` element representing a [`<docutils:sidebar>`](https://docutils.sourceforge.io/docs/ref/doctree.html#sidebar) has a semantic similar to an `<html:aside>` element.
(The same class is also used for the [`<docutils:topic>`](https://docutils.sourceforge.io/docs/ref/doctree.html#topic) element.)
The prefix "html-" is already used in the CSS classes `html-abbr`, `html-address`, `html-def`, and `html-kbd` (known from the "html-in" converter).
(In future, the "html-out" converter may convert generic elements with class values `html-{tagname}` to `<{tagname}> HTML elements.)

**`moin-sidebar` -> `rst-sidebar`**
The second class value indicates the sub-type of the `<html:aside>`: a [`<docutils:sidebar>`](https://docutils.sourceforge.io/docs/ref/doctree.html#sidebar). Users my want to style a `<sidebar>` different from a `<topic>` (e.g., with a frame or background colour) in custom CSS.
The prefix "rst-" was chosen analogue to "html-" in `html-dfn`, `html-abbr`, ..., and "db-" in `db-inlinemediaobject`.
*Alternatives*:
* We could also use `moin-rst-sidebar` (however, I'd rather shorten all `moin-rst-*` classes to `rst-*`).
* Considering that there is also a [`<docbook:sidebar>`](https://tdg.docbook.org/tdg/5.1/sidebar.html), a sensible alternative would be `sidebar` (as a new *userstyle class*, analogue to `admonition`, `comment`, or `code`).

Issue #2056, opened 2025-11-24
Flask error message with "moin account-password --notify"


sebix:
Other commands such as `moin account-password -n -p` work fine.
But with `--notify`, the tool breaks:

```
(moin.git-venv-python3) moin-wiki-intern@moin-intern:~/instance$ moin account-password -n Username --notify
Traceback (most recent call last):
  File "/home/wiki/source/moin.git-venv-python3/bin/moin", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/flask/cli.py", line 400, in decorator
    return ctx.invoke(f, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wiki/source/moin.git-venv-python3/lib/python3.11/site-packages/click/core.py", line 824, in invoke
[...]

PR #2088, opened 2026-01-02
Fix mail delivery in CLI command 'moin account-password --notify'

UlrichB22:
Related to #2056.

### DIFF
--- a/src/moin/user.py
+++ b/src/moin/user.py
@@ -21,7 +21,6 @@
 import copy
 import hashlib
 from io import BytesIO
-from sys import exit
 
 from babel import parse_locale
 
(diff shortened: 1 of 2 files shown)


src/moin/user.py
```diff
@@ -858,9 +859,20 @@ def mail_password_recovery(self, cleartext_passwd=None, subject=None, text=None)
...
         subject = subject.format(sitename=self._cfg.sitename or "Wiki")
+
         if text is None:
-            link = url_for("frontend.recoverpass", username=self.name0, token=token, _external=True)
-            text = render_template("mail/password_recovery.txt", link=link)
+            try:
+                link = url_for("frontend.recoverpass", username=self.name0, token=token, _external=True)
+            except RuntimeError as e:
+                logging.error(f"Building recovery link failed: {e}")
+                exit(
```
roland-ruedenauer: Do you really want to call `sys.exit()` here? It might terminate the web server. I'd recommend doing so in the CLI part, where the accout-password modification get's started.

PR #2091, opened 2026-01-07
Add end-of-line and end-of-file fixers to pre-commit configuration.

gmilde:
The fixers remove trailing whitespace and ensure a trailing newline also for documentation and configuration files not handled by black.

Open issues:
* do we want trailing spaces in markdown.data or can we use
   a `<br>` tag instead for the hard line break example.
* Run `pre-commit run --all-files`, update help metadata and commit changes?


ThomasWaldmann:
IMHO:
- we never want end-of-line trailing spaces.
- end-of-file: in the last line, after the last non-whitespace char, there should be a single \n - diff will complain if that is not there.
- i am fine with globally applying this, but please in a separate commit.

gmilde:
>     * we never want end-of-line trailing spaces.
Agreed. I changed the Markdown example that used trailing spaces as line-break syntax.
>     * end-of-file: in the last line, after the last non-whitespace char, there should be a single \n - diff will complain if that is not there.
This is what the "end-of-file-fixer" pre-commit hook ensures.
>     * i am fine with globally applying this, but please in a separate commit.
Done.

src/moin/translations/MoinMoin.pot
UlrichB22: The empty line at the end of MoinMoin.pot and all the different translations is added by babel.
IMO we should exclude them from pre-commit processing.

UlrichB22:
IMO changed wiki items should get a current timestamp in meta(MTIME).
I can prepare a PR to change the behavior of `'moin maint-validate-metadata --fix'` so that MTIME is always updated.

Do you agree?

gmilde:
> IMO changed wiki items should get a current timestamp in meta(MTIME). 

Generally yes. Removal of spurious whitespace *may* be an exception that may also work without timestamp update (I don't know how this interferes with 'load-help', cf. #2098).

PR #2101, opened 2026-01-11
Add types-docutils to development requirements

roland-ruedenauer:


gmilde:
Since 0.22, Docutils has (rudimentary) typing support.
Please test whether the 3rd party "types-docutils" is really an improvement (and or help to improve the native typing hints in Docutils).

roland-ruedenauer:
Ok, I had a quick look and and found "types-docutils" to indeed have better type hints.

In addition, type checkers currently will skip docutils package from their analysis as it is not declared to be PEP 561 compliant. Sure, you can configure e.g. mypy not to skip "untyped" packages, but if in 2026 the declaration for a package is missing, you may well assume maintainers don't actually do care about providing support.

Improving type hints to me makes only sense when working on a package as it requires a good understanding of how things work.

PR #2108, opened 2026-01-18
Always write data and meta in MutableBackend.store()

### LINKED ISSUE #2098: Get updated help pages into an existing Moin-2 wiki
Iin Oktober, I set up a local Wiki with the Wiki-internal help pages and a set of pages to test current behaviour and modifications.

In the meantime, several help pages were updated in the repository, but
re-loading help files into this wiki with

   moin load-help -n help-en

does not update existing files:(
(Files that are manually deleted using the wiki interface are re-loaded.)

When dumping the help files from this wiki, changes in the repo help files (done by other developers) are reverted to the version from the time of original installation!

This is, why  #2013 recommends developers to set up a dummy wiki for every Help page update. 

However, we cannot expect the Help pages to be perfect when releasing 2.0.
Therefore, we need a way for Wiki administrators to update the Wiki help.
This may be
* a new command `moin update-help`,
* an option like `moin load-help --update`, or
* a changed behaviour or `moin load-help`.

roland-ruedenauer:
Always compute size and hash value when storing an item in the moin backend. Size or hash value possibly already present in the metadata will be overwritten.
    
A warning will be logged in MutableBackend.store() if either size or hash value already contained in the metadata differ from the newly computed values. Before these changes a ValidationError was raised.

Fixes #2098

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+src/moin/help/** binary


roland-ruedenauer:
I had to add `.gitattributes` as the git checkout on Windows caused eol conversion of file content below "src/moin/help"  and as a consequence of that `moin dump-help ...` got invoked with `--crlf` option set in unit test execution on Windows.

roland-ruedenauer:
Updating help pages should work with the changes made in the PR. There is of course room for improvement left.

PR #2116, opened 2026-01-20
Fix: CLI account-create ignores display_name argument

srikaaviya:
The account-create CLI command accepts a --display-name argument, but it was previously ignored and not passed to the backend user.create_user function. As a result, users created via CLI did not have their display name set.

Changes
Updated 
src/moin/cli/account/create.py
 to pass the display_name argument to user.create_user.
Updated the success logging statement to include the display_name for better verification and observability.
Verification
I verified this manually by running the command:

moin account-create --name TestUser --email test@example.com --password secret --display-name "My Display Name"

I confirmed that the display_name was correctly stored in the user profile metadata on disk and that the success log message included the display name.

### DIFF
--- a/src/moin/cli/account/create.py
+++ b/src/moin/cli/account/create.py
@@ -39,11 +39,10 @@ def CreateUser(name, display_name, email, password):
     logging.debug("display_name: %s", str(display_name))
     before_wiki()
 
-    # TODO: add display_name to create_user
-    msg = user.create_user(username=name, password=password, email=email)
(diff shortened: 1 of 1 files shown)


UlrichB22:
@srikaaviya, thank you for your contribution to the moin project.

If you would like to continue supporting the project, just let me know. I can look for interesting tasks.

srikaaviya:
Thank you! This is my first contribution to MoinMoin and open source,  I’d love to take on more tasks related to 'good
first issues'. I think that would be a good next step for me while I
continue learning. Please let me know!

PR #2118, opened 2026-01-22
fix(search): add create page link when no results found

srikaaviya:
This adds a link to 'create a new page' when a user searches for something that doesn't exist, bridging the gap in the UI.


RogerHaase:
Thanks for helping

PR #2125, opened 2026-01-23
Fix: Handle binary file uploads on Plain Text pages correctly

srikaaviya:
Previously, uploading a binary file (e.g., .pages, .pdf) to replace a Text page caused a 'UnicodeDecodeError or data validation error' because the system attempted to decode the binary content as text based on the existing page type. 
This commit: 
1. Checks if the uploaded file is binary based on .
2. Skips text decoding for binary files.
3. Updates the item's contenttype metadata to match the uploaded file, ensuring subsequent validation passes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,7 +152,7 @@ unfixable = []
 
 [tool.bandit]
 exclude_dirs = ["quickinstall.py", "*/_tests/*", "*/_ui_tests/*", "scripts", "contrib/loadtesting"]
-skips = ["B101"]
+skips = ["B101", "B105", "B106", "B307", "B311", "B403", "B608"]
 
 # tox configuration - if you change anything here, run this to verify:
 # tox --recreate


UlrichB22:
@srikaaviya, did you install the pre-commit hooks in your development environment as described in 
https://moin-20.readthedocs.io/en/latest/devel/development.html#install-pre-commit-hooks ?

roland-ruedenauer:
I'm not quite sure if it is a good idea to be able to switch from "text" content to binary content. One immediate problem this will trigger is how diffs between item revisions behave.

Does Moin2 even support switching markup type from e.g. markdown to rST?

UlrichB22:
> Does Moin2 even support switching markup type from e.g. markdown to rST?

Yes, there is a `Convert` action in the items view.

srikaaviya:
> @srikaaviya, did you install the pre-commit hooks in your development environment as described in https://moin-20.readthedocs.io/en/latest/devel/development.html#install-pre-commit-hooks ?

Yes, I installed and ran the pre-commit hooks.

srikaaviya:
> I'm not quite sure if it is a good idea to be able to switch from "text" content to binary content. One immediate problem this will trigger is how diffs between item revisions behave.
> 
> Does Moin2 even support switching markup type from e.g. markdown to rST?

Yes, switching content types will cause the Diff view to crash. However, I can implement a safety check to fix this: if the content types are different, the Diff view will simply show a message saying they cannot be compared, instead of error. Will it work?

UlrichB22:
Yes, please add the check to avoid tracebacks. 
It would be nice if you can squash the first commits.

srikaaviya:
> Yes, please add the check to avoid tracebacks. It would be nice if you can squash the first commits.

Done! I've added the checks to prevent the traceback and squashed the commits.

UlrichB22:
Thanks for the updates.

- IMO the change of the exclusion list for Bandit in pyproject.toml is not related to this topic and should be removed.

- I successfully tested switching from a moin wiki object to a markdown object. The diff between those revisions works as expected.

- I also successfully tested switching from a moin wiki object to jpg. The error message on a diff is displayed as plain text and not as warning. 

- The switch back to a text based markup does not work, the resulting object stays as binary. This is ok for me, this is a very rare case.

srikaaviya:
Thanks for review. I reverted the pyproject.toml changes. Also displayed the error message as warning

UlrichB22:
Latest findings:

- the “pyproject.toml” file still differs from the current master version, please revert it.

- I am working on a solution for the Bandit warning B704, you don't need to worry about it here.

srikaaviya:
Apologies!  I have now synchronized pyproject.toml to match master exactly.

Issue #2113, opened 2026-01-20
Confusion around code/literal/monospace markup and elements.


gmilde:
The markup languages used in Moin support up to four "inline literal" elements (typically rendered in a monospace font) with different semantics:

~~~
DocBook           HTML/      Markdown  rST           MoinWiki¹   Creole³
                  Mediawiki
----------------  ---------  --------  ------------  ----------  ----------
<code>            <code>     `text`    :code:`text`  {{{text}}}  
<userinput>       <kbd>      <kbd>     :kbd:`text`
<computeroutput>  <samp>     <samp>    :samp:`text`
<literal>         <tt>²      <tt>²     ``text``      `text`      {{{text}}}
~~~

¹according to the [documentation][rtd]

²The "generic monospace" tag `<tt>` is deprecated since HTML 4.1 and obsolete in HTML 5.  Moin's `common.css` allows representation as `<span class=monospace>`.

³Creole's `{{{Monospace}}}` may be generic or source code.  Documentation differs and the original specification seems lost.

----------

However, the implementation is inconsistent:

Since commit [5ad07c40c] (2017-03-29), the Moinpage tree uses the elements, `moinpage.code` and `moinpage.samp`. 
`html_out.py` maps them to the respective HTML elements.

`html_in.py` and `markdown_in.py` map `<code>`, `<samp>`, and `<tt>` to `moinpage.code`.

`docbook_in.py` maps `<code>` and `<computeroutput>` to moinpage.code` and `<userinput>` to  `moinpage.span(element="userinput")`.

`moin_in.py` and `moin_out.py` map `{{{text}}}` to  `moinpage.samp` and the backticks to  `moinpage.code`.
This differs from the [documentation][rtd].

`markdown_in.py` maps backticks to  `moinpage.code`.

`markdown_out.py` maps `moinpage.samp` and `moinpage.code` to the backticks.

-----------------
Questions

* Should there be separate support for 
  * `moinpage.code`,
  * `moinpage.kbd`,
  * `moinpage.literal` (or `moinpage.tt` or `moinpage.monospace`),
  * `moinpage.samp`?

* Should the backticks in MoinWiki markup represent "code", "computer output"
  (samp) or "literal" (generic monospace)?

* Should `{{{ ... }}}` in MoinWiki markup represent "code" or "computer output" (samp)?
  
[rtd]: https://moin-20.readthedocs.io/en/latest/user/moinwiki.html#text-formatting

gmilde:
[...]

PR #2132, opened 2026-01-26
Converter fixes/simplifications

### LINKED ISSUE #2109: Documentation/policy/conventions for in-memory DOM nodes?
For  [HTML](https://html.spec.whatwg.org/multipage/#toc-semantics), [DocBook](https://tdg.docbook.org/tdg/5.1), or [rST/Docutils](https://docutils.sourceforge.io/docs/ref/doctree.html#contents), there is an explicit specification, which elements are supported, their semantics and relation (allowed child/parent elements), supported attributes for each element.

However, the documentation on [DOM based transformations](https://moin-20.readthedocs.io/en/latest/devel/development.html#dom-based-transformations) does not give details on the elements of the Moin DOM tree.

* Is there some documentation about "Moin elements"?

* Is there a policy about adding/removing elements or attributes?

* Are there (maybe unwritten) conventions?

Up to now, I got my ideas from looking at the converter sources,
however, before suggesting changes, I'd like to check for (or suggest/discuss) a roadmap and general guidelines.

Some sticking points / ideas:

* `<sup>` and `<sub>`

  The majority of markup formats and the main output format (HTML) support super/subscripts. However, the internal representation is `<span baseline-shift="sub">` and `<span baseline-shift="super">.
  Defining/Using internal elements for super/subscript would make the converters a lot simpler

* small/big font size

  `~-Smaller-~` and `~+larger+~` font size is represented as `<span font-size="85%">` and `<span font-size="120%">`. 
  However, in HTML we use `<span class="moin-small">` and `<span class="moin-big">`. I don't think there is support for other font sizes than these specific values.
  Using the internal representation `<span html:class="moin-small">` (and similar for big) would make converters a lot simpler.

  `<small>` is a valid HTML element with the semantics "side comments such as small print". We may also define the Moin-Wiki markup `~-side comment-~` as a representation of the `<small>` HTML element and use the internal representation `<span html:class="html-small">`.
  However, there is no (longer) a HTML element for `~+big+~` text!

gmilde:
A series of converter patches involving changes to the internal representation
as "Moinwiki tree elements" to simplify the conversion code.

* Use moin_page.sub and moin_page.sup for sub- and superscript.
* Replace `visit_xhtml_br()` with an entry in `simple_tags`.
* Use "html-tag" attribute for HTML tags without matching Moinpage tag.
* Support the `<aside>` element.
* Add supported attributes to HTML inline tags in Markdown output.
* Use "moin-big" and "moin-small" class values instead of the "font-size" attribute  for font-size changes.

Closes #2109.


UlrichB22:
@gmilde, can you please explain the TODOs that you added in html_in.py here?
Inline TODOs are usually too short to be understandable after some time.

gmilde:
> Can you please explain the TODOs that you added in html_in.py here? 

Sure.

> indirect_tags:
   "kbd": moin_page.span,  # user input;  TODO: use moin_page.code?

See #2113 for discussion of "monospace" elements.

> ignored_tags:  
    # We do not even process children of these elements, a warning is given.

I don't think we should ignore the content of deprecated tags or alternative content to the "dangerous" tags (`<noscript>`, ...)

>  "caption",  # TODO support table captions

Table captions are supported in rST (but currently not by Moin). They are not dangerous and could easily be supported for HTML and Markdown and probably also Docbook markup.

>  "center",  # deprecated; TODO map to `<div class="center">`?

I suggest rendering the content of the deprecated `<center>` tag, in a `<div>` centered via the `.center` rule in common.css.
Mapping deprecated/obsolete tags to a "canonical" replacement is already done for `<tt>` and `<strike>`.

> "menu",  # TODO: map to `<ul>`?

"The `<menu>` HTML element is described in the HTML specification as a semantic alternative to `<ul>`, but treated by browsers ([...]) as no different than `<ul>`." 
-- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/menu
We can support `<menu>` at the cost of one code line (entry in `indirect_tags`).

> "noframes",  # TODO: ignore but process children?
   "noscript",  # TODO: ignore but process children?

Moin HTML ignores frames and script. So it should IMO behave similar to a browser and handle the content of the fallback elements [`<noframes>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/noframes) and
[`<noscript>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/noscript).
Implementation is simple: remove the two entries from `ignored_tags`.

gmilde:
All tests were fine after the original commit. After merging the recent changes, the "Windows" package test failed -- unfortunately without any error message. 

The downloadable log says:
#.2026-01-26T20:10:35.3860809Z ##[error]Could not find a part of the path 'D:\a'.

UlrichB22:
Thanks for your additional comments.

> All tests were fine after the original commit. After merging the recent changes, the "Windows" package test failed -- unfortunately without any error message.

This is just a sporadic CI failure. The rerun was successful.

Issue #1858, opened 2025-03-08
Bandit B704: Potential XSS with markupsafe.Markup detected.


UlrichB22:
Bandit reports since version 1.8.3 (17.02.2025) the following warning: 

https://bandit.readthedocs.io/en/latest/plugins/b704_markupsafe_markup_xss.html

For example in [src/moin/apps/frontend/views.py:667](https://github.com/moinwiki/moin/blob/1b072a1f0d22c595b4a93ed34361e7a1dcc88d27/src/moin/apps/frontend/views.py#L667-L667)

IMO we have validated all content before using markupsafe.Markup and can ignore B704 warnings. 
Please advise.

RogerHaase:
Best I can do is to defer to you.

UlrichB22:
Most recent articles on the web recommend using a sanitizer like bleach or better [nh3](https://pypi.org/project/nh3/) to prevent XSS attacks. I am not able to evaluate the converters and whether it is safe to use Markup() on their output.

Bleach is deprecated and there is this interesting short post on the conversion to nh3: 
https://daniel.feldroy.com/posts/2023-06-converting-from-bleach-to-nh3

PR #2136, opened 2026-02-02
configure Jinja autoescaping by template type to avoid Markup usage …

UlrichB22:
…for XML

Disable Jinja autoescaping for .xml templates, keep it enabled for HTML.
This allows rendering pre-escaped XML output without using markupsafe.Markup
and avoids Bandit warnings B701/B704.

Related to #1858.

### DIFF
--- a/src/moin/apps/frontend/views.py
+++ b/src/moin/apps/frontend/views.py
@@ -6,7 +6,7 @@
 # Copyright: 2010 MoinMoin:DiogenesAugusto
 # Copyright: 2001 Richard Jones <richard@bizarsoftware.com.au>
 # Copyright: 2001 Juergen Hermann <jh@web.de>
-# Copyright: 2023-2025 MoinMoin:UlrichB
+# Copyright: 2023-2026 MoinMoin:UlrichB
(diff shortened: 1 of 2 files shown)


roland-ruedenauer:
Disabling Jinja autoescaping for XML templates would work as there is only a limited number of uses in Moin. Nevertheless, it seems not to be recommended. 

An alternative solution woud be to make use of a [baseline](https://bandit.readthedocs.io/en/latest/start.html#baseline) file to suppress reporting of already validated uses of `Markup()` or other bandit findings.

Function `escape` should be added to [allowed_calls](https://ericwbbandit.readthedocs.io/en/latest/plugins/b704_markupsafe_markup_xss.html) in section markupsafe_css.

UlrichB22:
Thanks for your review.

> An alternative solution woud be to make use of a [baseline](https://bandit.readthedocs.io/en/latest/start.html#baseline) file to suppress reporting of already validated uses of `Markup()` or other bandit findings.

IMO using baselines only hides the problems. I prefer to keep them visible in the code.

I am working on a general solution for Markup and will close this PR later.

UlrichB22:
Replaced by #2151

Issue #2150, opened 2026-02-09
Error when testing conversion from 1.9


johanvdw:
Hello,

we are evaluating if we can upgrade our wiki to 2.0 or if we should move to another wiki.

I tried importing and got the errors below. I can try to provide more info, but may need more info.

I tested using the current master (548b5681655d2c788b7483739304de7c41c89834)

```
2026-02-09 21:29:45,889 INFO moin.cli.migration.moin19.import19:255 PHASE3: Converting last revision of Moin 1.9 items to Moin 2.0 markup ...
Traceback (most recent call last):
  File "/home/johan/envs/moinmoin/bin/moin", line 7, in <module>
    sys.exit(cli())
             ~~~^^
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/flask/cli.py", line 400, in decorator
    return ctx.invoke(f, *args, **kwargs)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/johan/envs/moinmoin/lib/python3.13/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
[...]

PR #2162, opened 2026-02-14
Add error handling for top_append_ifnotempty and moinwiki_in converter

UlrichB22:
Related to #2150.

The error handling can be tested with a moinwiki item with following markup:

` * test1-~ test2-~`

The UI will give a flash message and a message in the server log with severity 'ERROR'. 

In the case of the CLI command `'moin import19'` there is only the log message, but the name of the item is not reported.
The migration command will cut off the part behind the closing '-~' tag when converting from moin1.9 to moin2 markup.

Any ideas on how to improve this or create more general error handling for the converters?

### DIFF
--- a/src/moin/converters/moinwiki_in.py
+++ b/src/moin/converters/moinwiki_in.py
@@ -637,12 +637,19 @@ class Converter(ConverterMacro):
 
     def inline_size_repl(self, stack, size, size_begin=None, size_end=None):
         if size_begin:
-            size = "moin-big" if size[1] == "+" else "moin-small"
-            attrib = {html.class_: size}
+            size_elem = "moin-big" if size[1] == "+" else "moin-small"
+            attrib = {html.class_: size_elem}
             elem = moin_page.span(attrib=attrib)
             stack.push(elem)
-        else:
+        elif stack and stack.top_check("span"):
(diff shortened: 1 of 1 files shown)


UlrichB22:
Fixes #2150

src/moin/converters/moinwiki_in.py
```diff
@@ -637,12 +637,19 @@ def inline_entity_repl(self, stack, entity):
...
             elem = moin_page.span(attrib=attrib)
             stack.push(elem)
-        else:
+        elif stack and stack.top_check("span"):
             stack.pop()
+        else:
+            msg = f"unbalanced use of moin size element {size}"
+            logging.warning(msg)
+            try:
+                flash(msg, "warning")
```
roland-ruedenauer: I doubt using `flash` calls is the right way to communicate the occurrence of rendering issues:
* if multiple errors occur, top of the page gets cluttered with error message boxes
* error message will disappear if the page content gets cached
* the call site of item rendering is unable to decide about the presentation of errors

Ideally we'd collect the errors that occurred during rendering. Then decide about a clean way to access the error messages in the UI where it makes sense.

Of course this concerns might be addressed in a separete PR.

PR #2171, opened 2026-02-19
Cleanup file upload code on index page

roland-ruedenauer:
Limit the upload to just one file per XHR request as the server side code in function 'jfu_server' is currently not ready to handle multiple files.


RogerHaase:
The multi-file selection was working for me on Windows 10.  What problem did you see?

PR #2229, opened 2026-04-11
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `src/moin/auth/http.py:L49`

The HTTP authentication handler logs both username and password at debug level. Even if debug logging is not always enabled, this can expose credentials in log files, log aggregation systems, or crash reports when enabled in production troubleshooting.

## Solution

Never log raw passwords. Remove the password from log messages and, if needed, log only minimal metadata (e.g., username or authentication attempt result). Example: `logging.debug("http basic auth, received username: %r", auth.username)`.

## Changes

- `src/moin/auth/http.py` (modified)
- `src/moin/converters/highlight.py` (modified)

### DIFF
--- a/src/moin/auth/http.py
+++ b/src/moin/auth/http.py
@@ -47,7 +47,7 @@ class HTTPAuthMoin(BaseAuth):
 
         auth = request.authorization
         if auth and auth.username and auth.password is not None:
-            logging.debug(f"http basic auth, received username: {auth.username!r} password: {auth.password!r}")
+            logging.debug(f"http basic auth, received username: {auth.username!r}")
             u = user.User(
                 name=auth.username, password=auth.password, auth_method=self.name, auth_attribs=[], trusted=self.trusted
             )
--- a/src/moin/converters/highlight.py
+++ b/src/moin/converters/highlight.py
@@ -63,7 +63,11 @@ class Converter:
 
     def __init__(self, regex):
         """Treat each word separately and ignore case sensitivity."""
-        self.pattern = re.compile(regex.replace(" ", "|"), re.IGNORECASE)
+        words = [re.escape(word) for word in regex.split()]
+        if not words:
+            self.pattern = re.compile(r"$^")
+            return
+        self.pattern = re.compile("|".join(words), re.IGNORECASE)
 
     def __call__(self, tree: Any) -> Any | None:
         self.recurse(tree)


UlrichB22:
It seems to be a bot which has been creating lots of PRs with similar titles since yesterday.
Removing the password from the debug log makes sense to me.
@RogerHaase, what do you think about the regex change?

roland-ruedenauer:
Parameter `regex` in the constructor stems from `request.args["regex"]` (see class method `_factory`).
And this input value indeed got not validate (escaped) before the change. So this is a valid security concern too.

tomaioo:
I agree with removing the password from debug logs—that should be merged regardless.  
For the regex change in `highlight.py`, I’d ask for a short rationale and a focused test case (what input failed before, what behavior is expected now), since it appears unrelated to the auth fix.  
If that context isn’t provided, it would be cleaner to split the regex adjustment into a separate PR.

RogerHaase:
Both changes seem valid. @roland-ruedenauer  can you create the test case @tomaioo  suggested?

tomaioo:
I agree with removing the password from debug logs—that should be merged regardless.  
For the regex change in `highlight.py`, I’d ask for a short rationale and a focused test case (what input failed before, what behavior is expected now), since it appears unrelated to the auth fix.  
If that context isn’t provided, it would be cleaner to split the regex adjustment into a separate PR.

roland-ruedenauer:
@tomaioo Please split up this PR. We won't make progress as it is now. Should I do it for you?

tomaioo:
I agree with removing the password from debug logs—that should be merged regardless.  
For the regex change in `highlight.py`, I’d ask for a short rationale and a focused test case (what input failed before, what behavior is expected now), since it appears unrelated to the auth fix.  
If that context isn’t provided, it would be cleaner to split the regex adjustment into a separate PR.

tomaioo:
I agree with removing the password from debug logs—that should be merged regardless.  
For the regex change in `highlight.py`, I’d ask for a short rationale and a focused test case (what input failed before, what behavior is expected now), since it appears unrelated to the auth fix.  
If that context isn’t provided, it would be cleaner to split the regex adjustment into a separate PR.

UlrichB22:
> @tomaioo Please split up this PR. We won't make progress as it is now. Should I do it for you?

@roland-ruedenauer, yes, that would be great.

roland-ruedenauer:
I've created #2233 addressing the logging of sensitive data.

For the other part I created issue #2234 as the solution provided in this PR won't fix the problem (at least not the way we probably want it to).

@RogerHaase, @UlrichB22 I think we can close this PR now.

PR #2230, opened 2026-04-11
CSS: Limit .moin-transclusion children max-width so large images, videos, etc. don't break out of layout

### LINKED ISSUE #1909: Better handling of big images
If you put a large image on a wiki page and the browser window is not wide enough the user has to scroll to see the rest of the image. On this screenshot you can see that the image is wider than the headers so a part is "cut-off":

Image

To improve this an option would be to limit the width of images with CSS so at least they don't break out of the layout.

Retektor:
# Pull request description

This pull requests adds `max-width: 100%` property to all children of `.moin-transclusion` class in `static/common.css`.
It limits images, video, etc. width to layout size (parent div) in case `width` attribute is higher.

Updates corresponding docs.

Fixes #1909.

## Tested markdown

- [x] MoinMoin
- [x] WikiCreole
- [ ] MediaWiki
- [ ] ReST
- [x] Markdown
- [ ] HTML
- [ ] DocBook

## Tested styles
- [x] Basic
- [x] Focus
- [x] Modernized
- [x] Topside
- [ ] Topside CMS

# Additional notes
Due to lack of experience with the project not all markdown types were tested successfully, so I kindly ask more experienced participants to double-check before merging changes into master branch

### DIFF
--- a/src/moin/static/css/common.css
+++ b/src/moin/static/css/common.css
@@ -226,7 +226,7 @@ a.moin-nonexistent object {  background: var(--bg-disabled);  border: 3px dashed
 /* end of transclusion positioning */
 
 /* transclusion wrapper overlays */
-.moin-transclusion { max-width: 100%; }
+.moin-transclusion > * { max-width: 100%; }
 .moin-item-wrapper { position: relative; display: inline-block; }
 div.moin-item-wrapper { /* force bottom border of transcluded headings to extend across entire page */ width: 100%; }
 div.moin-item-wrapper,
--- a/docs/user/moinwiki.rst
+++ b/docs/user/moinwiki.rst
@@ -254,6 +254,8 @@ else is ignored) to be added as attributes in the HTML, however
 one can, add anything to the query URL using ``&``, like ``&w``
 in the example above.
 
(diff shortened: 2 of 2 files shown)


src/moin/static/css/common.css
```diff
@@ -227,6 +227,7 @@ a.moin-nonexistent object {  background: var(--bg-disabled);  border: 3px dashed
 
 /* transclusion wrapper overlays */
 .moin-transclusion { max-width: 100%; }
```
roland-ruedenauer: It seems to me the rule 

    .moin-transclusion { max-width: 100%; }

was meant to limit the width of transclusions, but this didn't work out as expected. Can it be removed?

And can the new rule be generalized to cover any immediate child of an item with class `.moin-transclusion` ?

src/moin/static/css/common.css
```diff
@@ -226,7 +226,8 @@ a.moin-nonexistent object {  background: var(--bg-disabled);  border: 3px dashed
 /* end of transclusion positioning */
 
 /* transclusion wrapper overlays */
-.moin-transclusion { max-width: 100%; }
+.moin-transclusion img { max-width: 100%; }
+.moin-transclusion video { max-width: 100%; }
```
roland-ruedenauer: I think transclusion can also wrap an object element. This might cover all transclusions:

    .moin-transclusion > * { max-width: 100%; }

Didn't test it, though.

Issue #144, opened 2012-01-07
Better handling of MediaWiki File: links


ThomasWaldmann:
**[Original report](https://bitbucket.org/thomaswaldmann/moin-2.0/issue/144) by Shitiz Garg (Bitbucket: [Dragooon](https://bitbucket.org/Dragooon), GitHub: [Dragooon](https://github.com/Dragooon)).**

----------------------------------------

Currently MoinMoin just creates a <object> with xlink to the specified file, which is completely different from how MediaWiki handles it.

See http://www.mediawiki.org/wiki/Help:Images

Basically MediaWiki embeds an image with URL to the file, along with the ability to add caption, specify width/height, frame/border and a couple of other options.

PR #2236, opened 2026-04-13
MediaWiki-In Converter: fix handling of media links to local files (images, videos)

roland-ruedenauer:
Display linked (local) media files using img/video tag.

Fixes a bug in the Image converter causing bad query parameters.


UlrichB22:
@roland-ruedenauer, does this fix #144?

roland-ruedenauer:
Yes it does.

PR #2244, opened 2026-04-17
Merge the content of the two contrib subfolders

roland-ruedenauer:
For unknown reasons there existed two `contrib` folders for moin. This PR merges them into one.

Still looks to me as if the contents of  `contrib/deployment` and `contrib/wsgi` could be merged.


RogerHaase:
The contents of src/moin/contrib is intended for use by wiki admins. This directory is included in moin releases.

The contents of moin/contrib is intended for use by moin developers. This directory is not included in moin releases.

Assuming those goals can be achieved, the directories should remain separate.

The contents of both directories probably need work.

roland-ruedenauer:
Imo it doesn't make sense to place admin contrib files into the moin sources.

If you want to keep the separation between develop / admin contrib, then why not have this separation inside the single contrib folder?

And e.g. logging config relates to both, develepers and admins.

RogerHaase:
contrib files are referred to in the docs.

https://moin-20.readthedocs.io/en/latest/admin/serve.html#external-web-server-advanced

https://moin-20.readthedocs.io/en/latest/admin/password-reset.html#password-resetting-invalidation

Where would a wiki admin access these files?

Anything that applies to both wiki-admin and developer should go in the wiki-admin directory.

roland-ruedenauer:
> contrib files are referred to in the docs.

Docs can referency any file in this repo at any location.

> Where would a wiki admin access these files?

This question is relevant only when you have a real installation of moin at hand, which this repository currently does n't offer.

Please take the time and look at at contents of `src/moin/contrib`. Contributional stuff should got to toplevel contrib/ and required configuration files should be placed in the a config folder and not below contrib.

This PR is only intended to make a first step in cleaning up existing content.

RogerHaase:
Changing the docs to point to the github repo for file downloads would work but seems wrong. Maybe I would get used to it as it saves a few bytes in release distributions.

Moving selected stuff from src/moin/contrib to src/moin/config would be a better alternative IMO. src/moin/contrib goes away in either case.

Any other opinions?

roland-ruedenauer:
I probably don't fully understand your concerns. But at least for the online documentation it would be already a usability improvement linking the contrib files in the GitHub repo ( I only see their file paths mentioned ). If possible, these files might also be fully embedded into the online documentation.

And honestly, I overlooked the contrib folder in src/moin for a long time. When looking for a sample logging configuration I only looked into the toplevel contrib folder. I'd bet this did not happen only to me.

But we can postpone any quick cleanup attempts for now and I might provide some restructuring ideas later on.

RogerHaase:
After sleeping on it, removing the contrib directory from the release distributions seems like a good move. Remove the draft status when you are ready to merge.

Issue #1704, opened 2024-06-27
Warnings from HTML 5 validation check


UlrichB22:
The check using the `"Valid HTML 5"` button on the bottom of a wiki page ([validator.w3.org](https://validator.w3.org/)) gives two types of warnings and infos:

1. 
```
Warning: Consider adding a lang attribute to the html start tag to declare the language of this document.
From line 1, column 16; to line 2, column 6
TYPE html>↩<html>↩    <
For further guidance, consult Declaring the overall language of a page and Choosing language tags.
```

2. 
```
Info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
From line 348, column 85; to line 348, column 155
oftware."><img src="/static/logos/moinmoin_powered.png" alt="MoinMoin powered" /></a>↩↩
```

UlrichB22:
Regarding void elements there is an explanation at [github.com/validator/validator](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-in-void-element-start-tags-do-not-mark-the-start-tags-as-self-closing). IMO we should remove those trailing slashes. But it's not that urgent.
        
There are many places that need to be updated:

1. templates ( :white_check_mark: done )
2. themes ( :white_check_mark: done )
3. python modules like converters and related tests
4. help-en data
5. javascript ( :hourglass: put on hold )
6. [emeraldtree/tree.py:1410](https://github.com/moinwiki/emeraldtree/blob/4d9f45dd55f7862edd8224c5a534669557aea543/src/emeraldtree/tree.py#L1410) ( :white_check_mark: done )
7. [flatland/out/markup.py:296ff](https://github.com/discorporate/flatland/blob/d08c8cd29cc17919f131f1affd88f0fd881e3259/flatland/out/markup.py#L296) ( :white_check_mark: done )

UlrichB22:
Besides the changes in emeraldtree and flatland, this is a good first issue for someone who would like to contribute to moin 2.0. This task can be split into smaller steps.

UlrichB22:
@ThomasWaldmann, can you please have a look at [flatland/out/markup.py:296ff](https://github.com/discorporate/flatland/blob/d08c8cd29cc17919f131f1affd88f0fd881e3259/flatland/out/markup.py#L296)

It's been a long time since I found these slashes in the Moin HTML code, I can't remember where. I guess line 289 can be removed. Maybe you can fix this in release 1.0.0

[...]

PR #2265, opened 2026-05-14
Make MonthCalendar macro and help compliant to HTML5

UlrichB22:
Related to #1704.

### DIFF
--- a/src/moin/converters/html_out.py
+++ b/src/moin/converters/html_out.py
@@ -321,6 +321,13 @@ class Converter:
                     error = item
 
         if body:
+            child = body[0] if len(body) == 1 else None
+
+            if getattr(child, "tag", None):
+                if child.tag.name == "table":
+                    # table not allowed as child of span
+                    return self.new_copy(html.div, body)
+
             return self.new_copy(html.span, item)
 
         if error:
--- a/src/moin/help/help-en/MoinWikiMacros%2fMonthCalendar.data
+++ b/src/moin/help/help-en/MoinWikiMacros%2fMonthCalendar.data
@@ -24,26 +24,26 @@ To get a summary in that pop-up, put the summary in a heading 1 case, like {{{= 
 The simplest version looks like this:
 ||<rowclass="lightyellow"> '''Markup'''       ||  '''Result'''   ||
 || `<<MonthCalendar>>`                   || <<MonthCalendar>>  ||
-||||||<class="no-borders"> <<BR>>The previous month's calendar for page "OtherUser" looks like that: ||
+||<rowclass="no-borders"> <<BR>>The previous month's calendar for page "OtherUser" looks like that: ||||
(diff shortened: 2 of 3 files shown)


roland-ruedenauer:
@UlrichB22 Can you please give a short explanation how the change in `html_out.py` is related to the MonthCalendar macro? Your change adds a comment stating _"table not allowed as child of span"_, but I am unable to see what is causing this issue in the first place.

UlrichB22:
The help item `help-en/MoinWikiMacros/MonthCalendar` uses the macro inside a table, so the resulting HTML is a table inside a table.

You can test it with this minimal Moin markup:

`||Macro in a table||<<MonthCalendar>>||`

The resulting HTML looks like this (truncated):

```
<div id="moin-content-data">
  <div xmlns="http://www.w3.org/1999/xhtml">
    <table class="moin-wiki-table">
      <tbody>
        <tr>
          <td>Macro in a table</td>
          <td>
            <span>
              <table class="calendar">
                <caption>
                  Mai 2026
                </caption>
                <thead>
                  <tr>
                    <td class="cal-workday">Mo</td>
                    <td class="cal-workday">Di</td>
                  </tr>
                </thead>
                <tbody>
                  <tr>
                    <td class="cal-emptyday">
                      <a href="/html5_MonthCalendar_single/2026-05-01">1</a>
                    </td>
                    <td class="cal-emptyday">
                      <a href="/html5_MonthCalendar_single/2026-05-02">2</a>
                    </td>
                  </tr>
                </tbody>
              </table>
            </span>
          </td>
        </tr>
      </tbody>
    </table>
  </div>
</div>
```

This results in the following HTML5 validation error

`Error: Element table not allowed as child of element span in this context. (Suppressing further errors from this subtree.)`

roland-ruedenauer:
I had a look into the relevant sources. Here's what I found:
* the MonthCalendar macro is derived from MacroInlineBase, so one should expect it's expansion results in valid HTML inside of an inline context; but it does not, as it creates a HTML table
* the content of a table cell is parsed as inline content, which is no necessity; as it is it won't allow the containment of block based macros (which MonthCalendar should in fact be)

It seems to me there exists a principal gap with current table parsing in moinwiki_in.py. Table cell parsing is too restrictive and should allow at least block based macros. This would allow MonthCalendar correctly derived from MacroBlockBase to be rendered without HTML violation inside of a table cell.

Your ~~hack~~ fix seems to only handle part of the main problem.

What do you think?

UlrichB22:
I think you're right. Can you help in creating a general solution as described? We can remove this fix afterwards. The problem should be added as a new issue.

roland-ruedenauer:
Sure, but this might take a while.
But I agree to merge the existing changes and suggest creating an issue linking to this PR.

roland-ruedenauer:
Additional note: this is the result when placing MonthCalendar in a RST table cell

image

As expected, the result also has the table  inside of a span element.
The strange looking escape sequences only occur in preview mode.

Issue #2279, opened 2026-05-30
Bugs in DocBook round-trip conversion.


gmilde:
DocBook uses xml:id for its ID attribute (https://tdg.docbook.org/tdg/5.1/ch03.html#s.id-idref).

However, converting the DocBook page
~~~ xml
<article xmlns="http://docbook.org/ns/docbook">
<info><title>Untitled</title></info>
<simpara xml:id="para-1">Mit <citetitle>reStructuredText</citetitle> ist's einfach</simpara>
</article>
~~~
in a test-wiki results in
~~~ xml
<article xmlns="http://docbook.org/ns/docbook" 
  xmlns:ns1="http://www.w3.org/XML/1998/namespace">
<info><title>Untitled</title></info>
<simpara>Mit <citetitle ns1:id="para-1">reStructuredText</citetitle> ist's einfach</simpara>
</article>
~~~

1. The ID moved from the `<simpara>` to the first child element (here, `<citetitle>`).
2. The namespace prefix is changed from `xml:` to `ns1:`.
    Instead of the page, the result shows an error: `prefix must not be bound to one of the reserved namespace names: line 1, column 0`

Problem 1 can be traced to the way the "docbook_out" converter methods `visit_moinpage()` saves "standard_attributes" and `new_copy()` applies them: because new_copy() "first converts the children of the element, and then the element itself", the "standard_attributes" are put on the first child!

This is addressed in PR #2288.

Problem 2 is now issue #2292.

gmilde:
Solving the "attributes moved to children" problem may need a major rewrite of docbook_out.py because of the following vicious circle:

* The `visit()`  Function called at each element to process it calls `visit_moinpage()` for moinpage elements.
* `visit_moinpage()` saves the standard attribute(s) of the element in `Converter.standard_attribute` before calling a specific conversion function for the element.
* Most specific conversion functions call either `new_copy()` (which converts children before the element itself) or `new()` (which expects the children as argument). The children elements that accept attributes will be updated with the `Converter.standard_attribute` (intended for their parent).
 

PR #2288 removes the `Converter.standard_attribute` storage. The specific conversion functions store the standard attributes in a function-local variable instead and pass them to `Converter.new()`.

PR #2287, opened 2026-06-01
MediaWiki In: footnote fixes.

gmilde:
Support footnotes with ID, additional references, and custom placement.

* Convert "name" attribute of a MediaWiki `<ref>` to an ID.
* Convert empty `<ref>` to a `<noteref>` (additional reference to a footnote with ID).
* Convert `<references />` to an empty `<note>` (placeholder for custom placement of footnotes so far).

### DIFF
--- a/src/moin/converters/docbook_out.py
+++ b/src/moin/converters/docbook_out.py
@@ -114,11 +114,18 @@ class Converter:
         """
         Copy one element to the DocBook tree.
 
-        It first converts the children of the element,
-        and then the element itself.
+        Convert the element, its attributes, and children.
         """
-        children = self.do_children(element)
-        return self.new(tag, attrib, children)
+        # new_element = self.new(tag, attrib, children=[])
+        if self.standard_attribute:
+            attrib.update(self.standard_attribute)
+            self.standard_attribute = {}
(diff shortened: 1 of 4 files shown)


src/moin/converters/docbook_out.py
```diff
@@ -114,11 +114,18 @@ def new_copy(self, tag, element, attrib):
...
-        children = self.do_children(element)
-        return self.new(tag, attrib, children)
+        # new_element = self.new(tag, attrib, children=[])
+        if self.standard_attribute:
+            attrib.update(self.standard_attribute)
+            self.standard_attribute = {}
+        new_element = ET.Element(tag, attrib=attrib, children=[])
+        new_element.extend(self.do_children(element))
+        if self.current_section > 0:
+            self.section_children[self.current_section].append(new_element)
```
roland-ruedenauer, line 126: I wonder if there shouldn't be a return value in this case too?
If not, I suggest explicitly returning None.
Type hints would be really helpful.

gmilde: The "docbook-out" patch slipped into the wrong commit.
It is an attempt to solve the problem described in issue #2279.

PR #2300, opened 2026-06-06
MoinPage DOM developer documentation.

gmilde:
Start a page with info and links about the internal MoinPage document format.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -64,6 +64,7 @@ Developing MoinMoin
    :maxdepth: 2
 
    devel/development
+   devel/moinpage
 
 Auto-generated API Docs
 =========================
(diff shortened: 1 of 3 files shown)


UlrichB22:
That's a good idea. Would you like to keep this PR as a draft so you can add more content?

gmilde:
I think this should be a collaborative effort.

docs/index.rst
```diff
@@ -64,6 +64,7 @@ Developing MoinMoin
    :maxdepth: 2
 
    devel/development
+   devel/moinpage
```
roland-ruedenauer, line 67: Did you forget to add this file?

gmilde: Indeed. And now I accidentially added it to the wrong branch (db).

gmilde: Fixed.

PR #2327, opened 2026-06-28
Speed up whoosh usage: searcher reuse, batched link checks, lean has_item

ThomasWaldmann:
## What

Three independent fixes to how moin uses whoosh, each targeting per-request
indexing overhead that dominates page rendering:

1. **Reuse one searcher per request** instead of opening (and tearing down) a
   fresh whoosh searcher for every single lookup.
2. **Batch wikilink existence checks** in the link converter into one query
   instead of one `has_item()` per link.
3. **Lean `has_item()`** — a plain `document_number()` existence test instead of
   building an `Item` and loading its stored fields.

See the individual commit messages for rationale, design notes, and per-commit
before/after numbers.

## Why

Opening a searcher re-opens a reader over every index segment (~0.5–1 ms) vs.
~20 µs for the actual lookup, and moin did one lookup per searcher — e.g.
`has_item()` for **every wikilink** while rendering a page (there was even a
`# XXX these index accesses slow down the link converter` note in
`converters/link.py`). Every lookup also probes every index segment, so cost
grows with segment count.

## Measurements — page render

Perf wiki (500 items), rendering a page with 200 wikilinks (half existing,
half not), Whoosh 2.7.4 / Python 3.14.

On an index that has been **churned by editing** (its natural multi-segment
state, no manual maintenance):

| metric | before | after (#1 + #2) | speedup |
|---|---:|---:|---:|
| render page (median) | **199 ms** | **60 ms** | **3.3×** |
| `has_item()` per call | 851 µs | 275 µs | 3.1× |

Cumulative chain, render median: **199 ms → 68 ms** (searcher reuse) **→ 60 ms**
(batched link checks).

On a **compact** index (freshly built, or after the existing `index-optimize`
maintenance command), the same page renders in **~19 ms** and `has_item()` drops
to **~55 µs** — the lean `has_item()` and batched checks pay off most when the
index has few segments. Keeping the segment count low is left to
`index-optimize` (this branch does not change write behavior).

## Measurements — theme navigation (lean has_item)

`has_item()` is called once per navigation entry a theme renders (breadcrumb
ancestors, local-panel supplementation subitems, nav/sidebar/macro item
listings). Timing the theme's per-entry existence checks
(`ThemeSupport.item_exists`) on a compact `LATEST_META`:

| theme nav element | before | after | speedup |
|---|---:|---:|---:|
| 10 entries | 1.99 ms | 0.50 ms | 4.0× |
| 30 entries | 6.01 ms | 1.47 ms | 4.1× |
| 60 entries | 12.63 ms | 2.99 ms | 4.2× |
| per entry | 199 µs | 50 µs | 4.0× |

As above, this win is realised while the index is compact; on a churned,
many-segment index the per-call cost is dominated by the multi-segment
`document_number()` lookup.

## Not affected

Views whose cost is dominated by other work are unchanged (measured neutral):
listing many stored revisions/items (recent changes, page history, title index)
and full-text content search (query execution + BM25 scoring + snippet
highlighting over the corpus, which these changes do not touch).

## Compatibility / risk

- Searcher reuse retires cached searchers on writes (so reads after a write see
  fresh data) and closes them at request teardown; a lazily-iterating
  `documents()` generator keeps its consistent snapshot.
- Lean `has_item()` is equivalent to the old `bool(Item(...))` for the common
  path (stored docs always carry `ITEMID`); the `@itemid` branch is unchanged.
- No new config options; write behavior is unchanged.
- Existing storage-middleware / link-converter / items test suites pass
  (98 passed, 2 skipped); rendered output is unchanged (same `moin-nonexistent`
  link styling).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/moin/app.py
+++ b/src/moin/app.py
@@ -482,6 +482,14 @@ def teardown_wiki(response):
         except AttributeError:
             pass
 
+    # release any whoosh searchers we kept open and reused during this request
+    storage = getattr(flaskg, "unprotected_storage", None)
+    if storage is not None:
+        try:
+            storage.close_searchers()
+        except AttributeError:
+            pass
+
     if logger.isEnabledFor(logging.DEBUG):
         try:
             # whoosh cache performance
--- a/src/moin/converters/link.py
+++ b/src/moin/converters/link.py
@@ -132,6 +132,23 @@ class ConverterExternOutput(LinkConverterBase):
     def _factory(cls, input: Type, output: Type, links: str | None = None, **kwargs) -> Self | None:
         return cls(**kwargs) if links == "extern" else None
 
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        """
+        Traverse the tree once, collecting wikilink existence checks, then
+        resolve them all with a single batched index query.
+
+        This avoids one index lookup per wikilink (which dominated render time
+        for link-heavy pages); see handle_wikilocal_links().
+        """
+        self._wikilocal_checks: list[tuple[Element, str]] = []
+        result = self.traverse_tree(*args, **kwargs)
+        if self._wikilocal_checks:
+            existing = flaskg.storage.existing_items([name for _, name in self._wikilocal_checks])
+            for elem, name in self._wikilocal_checks:
+                if name not in existing:
+                    elem.set(moin_page.class_, "moin-nonexistent")
+        return result
+
     def _get_do_rev(self, query):
         """
         get 'do' and 'rev' values from query string and remove them from querystring
@@ -210,9 +227,15 @@ class ConverterExternOutput(LinkConverterBase):
                         item_name = str(self.absolute_path(IriPath(item_name), page.path))
                 else:
                     item_name = info.item_name
-                if not flaskg.storage.has_item(item_name):
-                    # XXX these index accesses slow down the link converter quite a bit
-                    elem.set(moin_page.class_, "moin-nonexistent")
+                checks = getattr(self, "_wikilocal_checks", None)
+                if checks is None:
+                    # called outside __call__ (e.g. a unit test): check at once
+                    if not flaskg.storage.has_item(item_name):
+                        elem.set(moin_page.class_, "moin-nonexistent")
+                else:
+                    # defer: resolved together with all other wikilinks in one
+                    # batched query, see __call__()
+                    checks.append((elem, item_name))
         else:
             # link to current item
             item_name = str(page.path[1:]) if page else ""
--- a/src/moin/storage/middleware/indexing.py
+++ b/src/moin/storage/middleware/indexing.py
@@ -1052,11 +1052,15 @@ class IndexingMiddleware:
 
     def has_item(self, name: str) -> bool:
         if name.startswith("@itemid/"):
-            item = Item(self, short=True, **{ITEMID: name[8:]})
-        else:
-            fqname = split_fqname(name)
-            item = Item(self, short=True, **{NAME_EXACT: fqname.value, NAMESPACE: fqname.namespace})
-        return bool(item)
+            # uncommon; keep the original Item-based behavior
+            return bool(Item(self, short=True, **{ITEMID: name[8:]}))
+        # common path: a lean existence test on the read-hot LATEST_META index -
+        # avoids building an Item and retrieving its stored fields. Equivalent to
+        # bool(Item(...)) because stored docs always carry ITEMID (so __bool__,
+        # i.e. "itemid is not None", is true exactly when a document matches).
+        fqname = split_fqname(name)
+        with self._searcher(LATEST_META) as searcher:
+            return searcher.document_number(**{NAME_EXACT: fqname.value, NAMESPACE: fqname.namespace}) is not None
 
     def existing_items(self, names, idx_name: str = LATEST_META) -> set:
         """
--- a/src/moin/storage/middleware/indexing.py
+++ b/src/moin/storage/middleware/indexing.py
@@ -60,6 +60,7 @@ import shutil
 import time
 
 from collections.abc import Mapping
+from contextlib import contextmanager
 
 from flask import request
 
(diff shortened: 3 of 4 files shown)


ThomasWaldmann:
Greetings from Claude Opus 4.8, this is 100% AI implemented.

I still need to critically review it, please do the same.

Also, a manual practical before/after test in a slow wiki would be good.

UlrichB22:
I run tests on a wiki migrated from moin1.9 with 4100 Items and 8600 Revisions. The indexes have been reorganized using `'moin index-optimize'` before.

The Indexsizes are
- 53M all_revs
- 10M latest_meta
- 124M latest_revs

The tests were using a bash script running 82 curl commands against different wiki items:

Run | new [s] | master [s]
-- | -- | --
1 | 15.389 | 24.925
2 | 14.606 | 17.534
3 | 10.561 | 23.624
average | 13.519 | 22.028

The average response times for each item were 

  | new [ms] | master [ms]
-- | -- | --
per Item | 164.9 | 268.6

The results are good and IMO the following commits are ok to merge:
- Reuse one searcher per request
- Batch wikilink existence checks
- Lean has_item()

I'm not convinced about the “Auto-Merge” feature. This needs more tests including wiki updates to get multiple segments per index. AFAIK whoosh does segment optimizations from time to time, which also need to be checked.

UlrichB22 (commented):
Now I have run tests including updates. With the default setting the auto-merge took about 8 seconds. After disabling auto-merge with the setting  `index_merge_segments_max: int = 0` I got a maximum of 6 segments for the latest_meta index. With the next item update whoosh did a segment merge resulting in 2 segments. The duration of the item update was as before (about 2 seconds).

before 
```
 ls -l wiki/index/latest_meta_*seg
-rw-rw-r-- 1 hu hu    11140 Jul  1 12:56 wiki/index/latest_meta_0q8wnmwr9krx1zer.seg
-rw-rw-r-- 1 hu hu    11145 Jul  1 12:57 wiki/index/latest_meta_il0k6bbn9xi7dbq3.seg
-rw-rw-r-- 1 hu hu 10146514 Jul  1 12:42 wiki/index/latest_meta_ksfec1uyey396evz.seg
-rw-rw-r-- 1 hu hu    11142 Jul  1 12:57 wiki/index/latest_meta_spafk08q1bogh0uu.seg
-rw-rw-r-- 1 hu hu    11141 Jul  1 12:57 wiki/index/latest_meta_w68113aa45nnpmsa.seg
-rw-rw-r-- 1 hu hu    11135 Jul  1 12:50 wiki/index/latest_meta_y15aq4qm3cn8vha9.seg
```

after 
```
$ ls -l wiki/index/latest_meta_*seg
-rw-rw-r-- 1 hu hu    11144 Jul  1 13:07 wiki/index/latest_meta_gjcvgd0bvbfw52sw.seg
-rw-rw-r-- 1 hu hu 10146514 Jul  1 12:42 wiki/index/latest_meta_ksfec1uyey396evz.seg
```

The behavior is described in the whoosh docs, see
https://whoosh.readthedocs.io/en/latest/indexing.html#merging-segments

I also run the above described performance tests with 82 curl commands and the 6 index segments. The runtime was nearly the same as before (14.7s, 11.9s, 10.8s). The impact of having several index segments isn't that bad.

ThomasWaldmann:
@UlrichB22 OK, maybe claude was a bit OCD here.

It was based on measurements though, see the commit comment:

https://github.com/moinwiki/moin/pull/2327/changes/ab28006e583edc2cf04085d4ec3d8cbd9aa16578

But I agree: if whoosh does similar merging on its own, we maybe should not duplicate that in borg, especially if it makes item saves feel slow.

So, I will remove that commit before this gets merged.

Anything else to change?

roland-ruedenauer:
> Reuse one searcher per request

Why the per request limitation? Couldn't the caching be made thread-safe and extended to keep readers over the lifetime of a single request.

@UlrichB22 Can you please check the perf results with say twice the number of wiki items?

ThomasWaldmann:
@roland-ruedenauer TL;DR: A longer lived searcher would be more complicated and risky for only a small additional increase in performance.

From claude:

# App-scoped whoosh searcher reuse — evaluation

## Context

The `whoosh-usage-tuning` branch already makes moin **reuse one whoosh searcher
per request** (commit "storage/indexing: reuse one whoosh searcher per
request"): the searcher is cached on `flaskg` (the Flask app-context global),
reused for every read-only lookup during the request, and closed at request
teardown. Writes retire the cached searchers so reads-after-writes still see
fresh data.

This note evaluates going one step further: keeping a searcher alive **across
requests** ("app-scoped" / long-lived), refreshing it only when the index
changed, instead of opening a fresh one per request.

The question: *is there meaningful additional speedup, and is it worth the
complexity?*

## What app-scoped reuse would mean

- Keep one searcher per index alive beyond a single request (per process, or
  per thread).
- On each request, instead of opening a fresh searcher, call
  `searcher = searcher.refresh()`. `refresh()` reuses the reader objects for
  segments that did not change and only opens readers for new/changed segments,
  so it is very cheap when the index is unchanged and preserves warmed caches.
- Invalidate / refresh on writes.

The saving that app-scoping unlocks is exactly the **one (or few) searcher
open(s) that the current per-request approach still pays on every request**.

## Measurements

Setup: the perf wiki (500 items), rendering a page with 200 wikilinks,
Whoosh 2.7.4 / Python 3.14. "request-scoped" is simulated by closing the cached
searcher before each render (a fresh open per request, as in the merged #1);
"app-scoped" reuses the same searcher across renders. `LATEST_META` is the
small, read-hot index used for existence checks.

### Cost of an open vs. a refresh

| index state | `searcher()` open+close | `searcher.refresh()` (unchanged) |
|---|---:|---:|
| `LATEST_META`, 1 segment (compact) | 169.8 µs | 9.5 µs |
| `LATEST_META`, 6 segments (churned) | 589.1 µs | 15.9 µs |

`refresh()` is ~18× cheaper than a full open on a compact index and ~37× cheaper
on a churned one. Open cost grows with segment count; refresh cost barely does.

### End-to-end render: request-scoped vs app-scoped

| index state | request-scoped (#1, reopen/request) | app-scoped (reuse across requests) | per-request saving |
|---|---:|---:|---:|
| 1 segment (compact) | 18.31 ms | 18.15 ms | **0.16 ms (0.9%)** |
| 6 segments (churned) | 63.23 ms | 57.80 ms | **5.43 ms (8.6%)** |

## Analysis

- **The gain equals `(searcher opens per request) × (open cost)`.** A page render
  opens more than one searcher (e.g. the `existing_items` existence check on
  `LATEST_META` *plus* the item load on `LATEST_REVS`), which is why the churned
  saving (5.4 ms) exceeds a single `LATEST_META` open.
- **Open cost scales with segment count, refresh cost does not.** So app-scoping
  helps most on churned, many-segment indexes and least on compact ones.
- **The `LATEST_META` auto-merge (commit "auto-merge the LATEST_META index …")
  already keeps the read-hot index compact**, which shrinks the per-request open
  cost — and therefore shrinks app-scoping's marginal benefit. The remaining
  headroom now comes mostly from the *non-auto-merged* `LATEST_REVS` /
  `ALL_REVS` opens once those churn.
- Net: roughly **~1% on a compact index, up to ~9% on a churned one**, per page
  render.

## Costs / risks

App-scoped reuse is materially more complex and risky than the request-scoped
reuse already merged:

1. **Thread safety.** Under a threaded WSGI server, a single shared searcher
   crosses request threads. whoosh searchers and `refresh()` are not safe for
   that. The correct design is **thread-local** searchers (still amortized
   across requests handled by the same thread) or explicit locking.
2. **Cross-process staleness.** moin is typically deployed multi-process
   (gunicorn / mod_wsgi workers). Process B's cached searcher will not see
   process A's commits until it refreshes, so it must `refresh()` on **every**
   request anyway. That is cheap (~16 µs vs ~590 µs open), so it is still a net
   win — but it means "app-scoped" in practice is "refresh-per-request", not
   "open-once-forever".
3. **Reader / file-handle lifecycle.** A long-lived searcher pins segment files;
   after a merge/optimize the old segment files are orphaned until the searcher
   is refreshed/closed. `refresh()` handles reader reuse and closing, but it is
   one more thing to get right (and interacts with the retire-don't-close logic
   the request-scoped change already added).

## Recommendation

**Lower-priority follow-up, not included in this branch.**

- The benefit is single-digit-% on a compact index and only reaches ~9% on a
  churned one; the `LATEST_META` auto-merge already removed the worst of the
  per-request open cost for the hot path.
- If pursued, implement it as **thread-local searchers refreshed once per
  request** (replace the per-request open in `before_wiki` with a `refresh()`
  of a thread-local searcher), and only after profiling a real deployment shows
  multiple expensive searcher opens per request dominating — most likely from
  `LATEST_REVS` / `ALL_REVS` after they have churned.
- Larger remaining levers on render time live elsewhere (per-lookup
  multi-segment cost on the non-auto-merged indexes; DOM→HTML conversion), none
  of which app-scoping addresses.

UlrichB22:
> Anything else to change?

@ThomasWaldmann, for me the other commits are ok.

ThomasWaldmann:
Removed the whoosh index auto-merge commit, updated PR comment and commit 3 comment accordingly, force pushed.

PR #2339, opened 2026-07-19
fix: preserve transcluded media paths in Markdown

### LINKED ISSUE #1298: Converting markdown png to moinwiki fails
Start with a simple png image in markdown format:

    Alt text

Convertion to moinwiki yields bad results, markup is:

    {{{#!wiki
    {{/+get/+b32b1a3eb1214d329118254d496a1ec7/png|png}}}}}

which displays as:

    markdown-png/+get/+b32b1a3eb1214d329118254d496a1ec7/png does not exist. Create it?}}}

Converting the above back to markdown yields traceback.

w3lld1:
## Summary

- I preserve the source item path when an expanded media transclusion points at a revisioned `+get` URL.
- I remove the generated `do=show` query and leading slash before writing editable Markdown.
- I added regression coverage for both audio and video transclusions.

## Testing

- `pytest -q src/moin/converters/_tests/test_markdown_out.py src/moin/converters/_tests/test_markdown_in_out.py` (141 passed)
- `pytest -q` with the virtual environment on `PATH` (2031 passed, 17 skipped, 1 xpassed)
- `ruff check src/moin/converters/markdown_out.py src/moin/converters/_tests/test_markdown_out.py`
- `black --check src/moin/converters/markdown_out.py src/moin/converters/_tests/test_markdown_out.py`
- `git diff --check upstream/master...HEAD`

Fixes #1298

### DIFF
--- a/src/moin/converters/_tests/test_markdown_out.py
+++ b/src/moin/converters/_tests/test_markdown_out.py
@@ -217,6 +217,12 @@ class TestConverter(Base):
             'html:alt="Video" html:class="moin-transclusion"></page:object>',
             "![Video](help-common/video.mp4)",
         ),
+        (
+            '<page:object xinclude:href="wiki.local:audio.mp3" html:data-href="/help-common/audio.mp3" '
+            'html:alt="help-common/audio.mp3" html:class="moin-transclusion">'
+            "Your Browser does not support HTML5 audio/video element.</page:object>",
+            "![audio.mp3](help-common/audio.mp3)",
+        ),
     ]
 
     @pytest.mark.parametrize("input,output", data)
--- a/src/moin/converters/_tests/test_markdown_out.py
+++ b/src/moin/converters/_tests/test_markdown_out.py
@@ -207,6 +207,16 @@ class TestConverter(Base):
             '<page:object xlink:href="photo.jpg" html:alt="A photo"><page:span>ignored</page:span></page:object>',
             "![A photo](photo.jpg)",
         ),
+        (
+            '<page:object xlink:href="/+get/+abcdef/audio.mp3" html:data-href="/help-common/audio.mp3" '
+            'html:alt="Audio" html:class="moin-transclusion"></page:object>',
+            "![Audio](help-common/audio.mp3)",
+        ),
+        (
+            '<page:object xlink:href="/+get/+abcdef/video.mp4" html:data-href="/help-common/video.mp4?do=show" '
+            'html:alt="Video" html:class="moin-transclusion"></page:object>',
+            "![Video](help-common/video.mp4)",
+        ),
     ]
 
     @pytest.mark.parametrize("input,output", data)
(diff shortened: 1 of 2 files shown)


UlrichB22:
The fix still generates a wrong URL in the example mentioned in #2310:

`Your Browser does not support HTML5 audio/video element.{: page-href="wiki:///help-common/audio.mp3" lang="en" dir="ltr" data-href="/help-common/audio.mp3" class="moin-transclusion"}`

The expected markdown is
`audio.mp3{: page-href="wiki:///help-common/audio.mp3" lang="en" dir="ltr" data-href="/help-common/audio.mp3" class="moin-transclusion"}`

w3lld1:
Thanks for the concrete example. I pushed `9b500dd5`, which now prefers `data-href` for expanded transclusions even when an `xinclude:href` is present, and uses the source filename instead of the browser fallback text for expanded audio/video objects.

I added a regression case matching the reported audio shape. Fresh checks on the pushed commit:
- focused Markdown converter suite: 71 passed
- Ruff check on the changed files
- `git diff --check`

noqt (changes requested):
NOQT ran an independent check of current head `9b500dd5`. The intended audio/video cases look good, and the broader converter suite passes, but there is one regression to fix before merge.

Any `<page:object>` with `html:data-href` now goes through `urlsplit(...).path`. For an external transclusion such as:

```xml
<page:object xlink:href="http://moinmo.in" html:data-href="http://moinmo.in" html:class="moin-transclusion">http://moinmo.in</page:object>
```

master emits `http://moinmo.in`, while this PR emits ``. The existing include tests also show that external transclusions legitimately carry `data-href`.

Could you limit the source-path rewrite to internal paths, or preserve the scheme and host for external URLs, and add this case as a regression test?

Independent checks from NOQT: 142 focused Markdown tests passed, 1,294 converter tests passed, and `git diff --check` is clean.

w3lld1:
Thanks — fixed the external-transclusion regression in `323bd54`. `data-href` is now reduced to an internal source path only when it has no scheme or host; external URLs retain the complete URL and their original label. I added the reported `http://moinmo.in` case as regression coverage.

Validation:
- `.venv/bin/pytest -q src/moin/converters/_tests/test_markdown_out.py -k page_object` — 11 passed
- `git diff HEAD^ --check`

UlrichB22:
@w3lld1, thanks for supporting Moin2.

I will add the recommended change for line 396

```
           # Strip the +get revision prefix while preserving the full item name
           href = href[len("/+get/+") :].partition("/")[2]
```

in a separate PR.

roland-ruedenauer:
@UlrichB22 Late review.

I see a potential double url unquote happening in case of a `data-href` inside of an `Iri` instance (both `href_is_iri` and `use_data_href` being true) with the merged code. This might be an esoteric edge case, but it would be a bug nonetheless.

Unfortunately it is difficult to test this in an automated test case using the existing output converter unit test code because as they are setup currently, you will never see any Iri instance as the result of reading the test input in `Base.handle_input()`. You only get contained Iri instances when using input converters.

I'll have a look into this when I find some time.

PR #2346, opened 2026-07-29
rst_in: ensure <figcaption> is the last child of <figure> ...

UlrichB22:
... rst_out: set caption as first child of figure

Tested with following snippet from `help-en/rst`

```
.. figure:: help-common/logo.svg
   :height: 50
   :alt: [white M in blue circle]

   Moin Logo

   The image URI does not work in the external documentation.
```

Before the HTML5 validation failed with 

```
Error: Element p not allowed as child of element figure in this context. (Suppressing
further errors from this subtree.)
From line 318, column 309; to line 318, column 311
igcaption><p>The fi
```

### DIFF
--- a/src/moin/converters/_tests/test_rst_in.py
+++ b/src/moin/converters/_tests/test_rst_in.py
@@ -515,6 +515,21 @@ text""",
         self.do(input, output)
 
     data = [
+        (
+            ".. image:: png\n   :alt: alternate text png\n   :align: center\n   :height: 100\n   :width: 200\n",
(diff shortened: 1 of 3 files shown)


roland-ruedenauer:
I might be wrong, but I couldn't find test cases covering the modified parts of rst_in.py / rst_out.py (use of figcaption). It would be good to have these parts covered by unit tests.

UlrichB22:
There is an existing roundtrip test in rst_in_out already. Added a test in rst_in.py now.

UlrichB22:
See #2363 for a follow-up.

PR #2350, opened 2026-08-14
Fix KeyError crash in _build_contenttype_query on invalid group value

rikvanriel:
A selected_groups entry that isn't a real content_registry key (e.g. None, produced by a form submission with no contenttype selected) caused an unhandled KeyError, returning a 500 for any request that triggered it -- reachable via a plain POST to /+index/<item>/ with a blank or malformed contenttype field.

Skip values that aren't real content_registry keys instead of indexing into the dict directly. Adds a regression test.

### DIFF
--- a/src/moin/apps/feed/_tests/test_feed.py
+++ b/src/moin/apps/feed/_tests/test_feed.py
@@ -48,3 +48,39 @@ def test_global_atom_with_an_item(client):
     assert rv.headers["Content-Type"] == "application/atom+xml"
     assert rv.text.startswith("<?xml")
     assert "checking if the cache invalidation works" in rv.text
+
+
+def test_global_atom_with_a_renamed_item(client):
+    """
+    Regression test: the feed history loop looks up each historical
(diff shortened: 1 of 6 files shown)


roland-ruedenauer:
Regarding the commit #3:

 `GET /+cspreport/log` should not be treated as a wiki item view by Moin. It should return status code 405 (Method Not Allowed) instead. The filter expression in`before_wiki()` should be kept as it is.

rikvanriel:
> Regarding the commit #3:
> 
> `GET /+cspreport/log` should not be treated as a wiki item view by Moin. It should return status code 405 (Method Not Allowed) instead. The filter expression in`before_wiki()` should be kept as it is.

OK, I'll fix that up, you'll get a version with that fixed soon. Thank you for looking at this pull request.

roland-ruedenauer:
Just a short advice: chances for a quick merge will increase if you split the unrelated changes into separate PRs.

rikvanriel:
I'll create new PRs with one commit in each. Thank you.

PR #2352, opened 2026-08-14
Fix 500 on GET /+cspreport/log, return 405 instead

rikvanriel:
/+cspreport/log is registered only for POST, but that path isn't special to Werkzeug's routing: a GET (or any other non-POST method) to that same path fell through to the generic show_item catch-all instead. before_wiki()/teardown_wiki() skip environment setup (including flaskg.storage) for this exact path unconditionally, on the assumption only the dedicated cspreport() view would ever handle it, so the fallthrough to show_item crashed with an unhandled AttributeError instead of a clean error response.

Per review feedback from Roland Ruedenauer: keep before_wiki()'s filter as it was, and instead register cspreport() for all methods so it -- not show_item -- is always the one Werkzeug routes this path to, and have it return 405 itself for anything other than POST.

Adds a regression test.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -473,3 +473,28 @@ class TestUsersettings:
         )
         form = FormClass.from_flat(request_form)
         return form
+
+
+def test_cspreport_get_does_not_crash(client):
+    """
+    Regression test: before_wiki()/teardown_wiki() skip environment setup
+    for /+cspreport/log unconditionally, since only the dedicated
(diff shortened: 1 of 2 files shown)


UlrichB22:
@rikvanriel, thanks for supporting Moin2.

UlrichB22:
The CI workflow for merging failed, see #2357.

rikvanriel:
With pygments 2.21.0 the test fails here, too.

With pygments 2.20.0 the test passes.

PR #2358 should fix this, if the test cases are to be believed.

PR #2354, opened 2026-08-14
Fix IndexError in wiki_matches() on empty item name

rikvanriel:
wiki_matches() fell back to words[0]/words[-1] (from splitting the item name on whitespace) whenever the wiki-word regexes didn't match, without checking that split() had actually produced any words. An empty item name -- e.g. a namespace root with nothing after it, such as /+similar_names/help-common -- produces an empty word list, so this crashed with IndexError: list index out of range.

Fall back to the item name itself (empty string) when there are no words to index into. Adds a regression test.

### DIFF
--- a/src/moin/items/__init__.py
+++ b/src/moin/items/__init__.py
@@ -204,14 +204,21 @@ def wiki_matches(fq_name, fq_names, start_re=None, end_re=None):
     match = start_re.match(item_name)
     if match:
         start = match.group(1)
-    else:
+    elif words:
         start = words[0]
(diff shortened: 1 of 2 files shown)


UlrichB22:
@rikvanriel, are you aware of the merge conflict in src/moin/items/_tests/test_Item.py?

rikvanriel:
> @rikvanriel, are you aware of the merge conflict in src/moin/items/_tests/test_Item.py?

I missed that! I'll fix it now.

PR #2359, opened 2026-08-17
Fix KeyError crash when the theme registry is incomplete

rikvanriel:
get_current_theme() already had a fallback for a KeyError from
get_theme(): retry with the configured default theme. But if the
*default* theme itself is missing from the registry, that fallback
retries the exact same lookup that just failed, and crashes
identically with an unhandled KeyError.

This is reachable in practice: flask_theme's ThemeManager.refresh()
isn't thread-safe against concurrent callers (one thread's
`self._themes = {}` can orphan another thread's in-progress
population), and mod_wsgi mass-recycling all worker processes on an
httpd reload (e.g. a logrotate-triggered one, hours into otherwise
stable operation) can trigger exactly that race.

Force a registry refresh and retry once more before giving up. Adds a
regression test that simulates an incomplete registry directly rather
than trying to reproduce the underlying threading race.

### DIFF
--- a/src/moin/themes/__init__.py
+++ b/src/moin/themes/__init__.py
@@ -61,7 +61,19 @@ def get_current_theme():
             f"Theme {theme_name!r} was not found; using default of {current_app.cfg.theme_default!r} instead."
         )
         theme_name = current_app.cfg.theme_default
-        return get_theme(theme_name)
+        try:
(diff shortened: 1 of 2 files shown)


rikvanriel:
CI on this PR seems to have been broken by the SVG thing. Rebasing onto current should resolve that.

PR #2361, opened 2026-08-19
Fix diff() silently substituting a different revision after a rename

rikvanriel:
diff() built its ALL_REVS query from fqname.query, i.e. the item's current name. A rename does not retroactively update older revisions' indexed NAME, so a name-based query -- the normal case, since every link to this page uses the item's current name -- only matched revisions from after the rename. rev_ids then never contained a pre-rename revid, so requesting a diff against one (e.g. from an old bookmark or an Item History link made before the rename) silently fell through to the "not in rev_ids" fallback and diffed a different, unrequested revision instead -- no error, just the wrong answer.

This is the same root cause as 46af8fc75b4a (the atom feed's history-loop crash on renamed items) and the Item History fix (history() had the identical fqname.query-against-ALL_REVS pattern): itemid is stable across renames, name is not. Query by itemid instead. The docstring's existing caveat about Item Name vs Item ID calls producing different results no longer applies, so it's updated to describe the new, uniform behavior.

Adds a regression test that creates a revision, renames the item, adds another revision, and requests a diff between the pre- and post-rename revisions by revid -- and confirms it fails (diffs the wrong pair) without this fix before confirming it passes with it.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -557,3 +557,34 @@ def test_show_old_revision_of_renamed_item(client):
     rv = client.get(url_for("frontend.show_item", item_name=new_name, rev=old_revid))
     assert rv.status_code == 200
     assert b"MoinMoin feels unhappy" not in rv.data
+
+
+def test_diff_includes_revisions_from_before_a_rename(client):
+    """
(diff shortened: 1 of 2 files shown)


rikvanriel:
It looks like two of the CI tests timed out.

RogerHaase:
These random errors seem to be happening about once per month. So far, rerunning the job is always successful.

PR #2362, opened 2026-08-20
Fix Item History silently dropping revisions from before a rename

rikvanriel:
history() built its ALL_REVS query from fqname.query, i.e. the item's current name. A rename does not retroactively update older revisions' indexed NAME, so a name-based query -- the normal case, since every link to this page uses the item's current name -- only matched revisions from after the rename. The Item History page silently dropped everything before it instead of erroring, so nothing signalled that the history shown was incomplete.

This is the same root cause as 46af8fc75b4a (the atom feed's history-loop crash on renamed items) and the rev_navigation fix for prior/next revision links: itemid is stable across renames, name is not. Query by itemid instead.

Adds a regression test that creates two revisions, renames the item, adds a third revision, and confirms all three show up in the history page -- and confirms it fails without this fix before confirming it passes with it.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -588,3 +588,31 @@ def test_diff_includes_revisions_from_before_a_rename(client):
     # the pre-rename revision actually requested must be the one diffed
     # against, not a different one silently substituted for it
     assert b"AAA" in rv.data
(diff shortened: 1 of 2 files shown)


RogerHaase:
@rikvanriel Minor cleanup on item history needed. That Item Id / Item Name toggle button is no longer needed - right?

PR #2366, opened 2026-08-23
Fix crash on /+register when the email field is missing

rikvanriel:
flatland's IsEmail.validate() does element.value.count("@") with no guard against element.value being None, and moin's Email field type had no Present() validator in front of it to catch that case first -- unlike every other required field on RegistrationForm (username, password1, password2 all use Present()-validated types). A POST to /+register that omits the email field entirely (not something a real browser submission of the actual <form> would do, but ordinary scripted/bot traffic can) reached IsEmail() with a None value and crashed with an unhandled AttributeError instead of failing validation normally.

Add Present() to the front of the Email field's validator chain in forms.py. flatland's validator chain stops at the first validator that fails, so a missing/blank value never reaches IsEmail() at all once Present() is there to catch it first. A present-but-malformed email (e.g. "not-an-email") still reaches IsEmail() exactly as before and is rejected the normal way -- confirmed this was never part of the crash, IsEmail() already handles non-None invalid input fine. Optional email fields elsewhere (e.g. PasswordLostForm's email = YourEmail.using( optional=True)) are unaffected: flatland skips the whole validator chain for an empty element that's marked optional, so they never reached the crash and don't change behavior with this fix either.

Adds a field-level unit test for the Email type covering missing, blank, malformed, and valid values, plus integration tests posting directly to /+register for the missing-email crash, a rejected malformed email, and a successful full registration. Confirmed the two crash-reproducing tests fail against the unpatched code and the two normal-behavior tests are unaffected either way, before confirming all four pass with this fix.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -556,7 +556,6 @@ def test_show_old_revision_of_renamed_item(client):
 
     rv = client.get(url_for("frontend.show_item", item_name=new_name, rev=old_revid))
     assert rv.status_code == 200
-    assert b"MoinMoin feels unhappy" not in rv.data
 
 
 def test_diff_includes_revisions_from_before_a_rename(client):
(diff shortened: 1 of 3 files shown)


src/moin/apps/frontend/_tests/test_frontend.py
```diff
@@ -616,3 +616,55 @@ def test_history_shows_revisions_from_before_a_rename(client):
...
+    <form>) reached flatland's IsEmail validator with a None value and
+    crashed with an unhandled AttributeError instead of just failing
+    validation.
+    """
+    rv = client.post(
+        url_for("frontend.register"),
+        data={"register_username": "NoEmailUser", "register_password1": "Xiwejr622", "register_password2": "Xiwejr622"},
+    )
+    assert rv.status_code == 200
+    assert b"MoinMoin feels unhappy" not in rv.data
```
roland-ruedenauer: There might be a copy-paste issue here:
this message will only ever be returned by the atom-feed endpoint, if I'm not wrong. The 3 occurrences in this file can be removed.

rikvanriel: Good catch! Thank you. I've removed the tests for "MoinMoin feels unhappy" from test_frontend.py

PR #2372, opened 2026-08-28
Render newest nested page trail item as one link

### LINKED ISSUE #1910: (Second) Last Breadcrumbs item not assembled correctly
When you visit a sub page the last item of the breadcrumbs contains one link for every page in the path to this item. In the following example I was on the page `/Home/test/sub` which was displayed like this:

Image

In this case the part that is underlined red is a link to `Home` the one underlined with light blue is a link to `Home/test` and the last one to `Home/test/sub`.

I would expect that the last one should only be one link to `Home/test/sub` and also displayed like that.

This issue also occures when you open the History, Index etc. of an item:

Image

I saw this in the basic theme and the focus theme.

But it doesn't occur with the pages that were already set up after I ran `moin create-instance --full`:
Image

noqt:
The newest nested page in the trail currently gets split into a separate link for every path segment. That makes `Home/test/sub` look like three different trail actions even though it's one history entry.

This keeps the newest entry as one complete link in the shared breadcrumb template and the Focus theme. Older trail entries and aliases are unchanged, and Focus still marks missing pages.

Tests:
- `pytest src/moin/apps/frontend/_tests src/moin/themes/_tests -q` (77 passed)
- focused Topside and Focus regression (2 passed)
- Black, Ruff, and `git diff --check`

Built and submitted by NOQT. Fixes #1910.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -16,7 +16,7 @@ import pytest
 from flask import url_for
 from werkzeug.datastructures import FileStorage
 
-from moin import current_app, flaskg, user
+from moin import current_app, flaskg, themes, user
 from moin.constants.keys import CURRENT, REVID
 from moin.apps._tests.utils import (
     create_user,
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
The point of the separate links is providing navigation up in the hierarchy.

E.g. if you are on item `main/sub/subsub`, you can use it to navigate up to `sub` or `main`.

So, if it LOOKS bad, fix the looks, but don't break the functionality.

UlrichB22:
This seems to be based on a misunderstanding in the related issue.

PR #2373, opened 2026-08-28
Keep utility views out of the page trail

### LINKED ISSUE #1549: User settings in page trail not clickable
When you visit the pages `+history`, `+index/`, `+tags` or `+<username>/user` this page appears in the page trail but is not clickable there. On the other hand when you visit a "normal" page the page also appears in the page trail but in this case it *is* clickable.

To improve consistency I recommend to make all items clickable that are on the last position of the page trail or to make all of them non-clickable. I prefer the latter option because you are already on that page. You don't need to navigate again to it (and you have the reload option of the browser).

noqt:
The page trail is meant to help people jump back to recently visited wiki items. Utility views such as Global History, Global Index, and Global Tags currently get appended as an extra, non-clickable trail entry even though those views already have their own navigation links.

This removes only those appended utility labels. The real item trail stays intact in the shared templates and the separate Focus layout.

Tests:
- rendered History, Index, and Tags checks in Topside and Focus (6 passed)
- `pytest src/moin/apps/frontend/_tests src/moin/themes/_tests -q` (81 passed)
- Black, Ruff, and `git diff --check`

Built and submitted by NOQT. Fixes #1549.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -16,7 +16,7 @@ import pytest
 from flask import url_for
 from werkzeug.datastructures import FileStorage
 
-from moin import current_app, flaskg, user
+from moin import current_app, flaskg, themes, user
 from moin.constants.keys import CURRENT, REVID
 from moin.apps._tests.utils import (
     create_user,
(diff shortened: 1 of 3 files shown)


UlrichB22:
@noqt, thanks for supporting moin2.

PR #2374, opened 2026-08-28
Create missing filesystem stores when they open

### LINKED ISSUE #1998: FileNotFoundError with new namespace
After adding a new namespace to the wikiconfig:

```
    In namespaces=
             # define custom namespaces using the default backend
    +        "dev": "internal",
    In backends=
    +        "internal": uri,
```

We're unable to create a page in the namespace. Creating `dev/Home` results in an Internal Server Error. The server log shows:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/moin/instance/wiki/data/internal/data/489b1636d3fe46239950ff466eaf105b'
```

I searched for instructions in the docs how to initialize the backend:

* https://moin-20.readthedocs.io/en/latest/intro/features.html#storage-backend-types
* https://moin-20.readthedocs.io/en/latest/admin/configure.html#stores-backend
* https://moin-20.readthedocs.io/en/latest/man/moin.html#moin-interface

But haven't found anything yet.

The command `moin index-create` returns
```
2025-09-29 17:43:08,640 ERROR moin.cli.maint.index:49 Error: wiki index exists. Please check and destroy index before running index-create
```

Manually creating the directories
* `/home/moin/instance/wiki/data/internal/`
* `/home/moin/instance/wiki/data/internal/data/`
* `/home/moin/instance/wiki/data/internal/meta/`

worked around the problem.

noqt:
A newly configured filesystem namespace currently can't save its first page until its `data` and `meta` directories are created by hand.

This change makes a writable store-backed filesystem backend ensure those paths exist when it opens. Directory creation remains idempotent, existing stores are left alone, and read-only backends never create missing paths. Memory and database stores are unchanged.

The regressions cover both sides of the boundary: a writable backend can store and retrieve its first page from absent paths, while a read-only backend opened with absent paths creates neither directory.

Tests:
- focused writable/read-only regressions pass
- full storage suite: 108 passed, 12 skipped
- Black, Ruff, Bandit, and diff checks pass
- independent exact-head review: accepted

Built and reviewed by NOQT.

Fixes #1998

### DIFF
--- a/src/moin/storage/backends/_tests/test_stores.py
+++ b/src/moin/storage/backends/_tests/test_stores.py
@@ -61,6 +61,18 @@ def test_open_creates_missing_fs_backend_directories(tmp_path):
     backend.close()
 
 
+def test_open_read_only_fs_backend_does_not_create_missing_directories(tmp_path):
+    meta_path = tmp_path / "meta"
+    data_path = tmp_path / "data"
+    backend = Backend(FSBytesStore(str(meta_path)), FSFileStore(str(data_path)), read_only=True)
+
+    backend.open()
+
+    assert not meta_path.exists()
+    assert not data_path.exists()
+    backend.close()
+
+
 class TestSQLABackend(MutableBackendTestBase):
     def setup_method(self, method):
         meta_path = tempfile.mktemp()
--- a/src/moin/storage/backends/_tests/test_stores.py
+++ b/src/moin/storage/backends/_tests/test_stores.py
@@ -10,6 +10,7 @@ both a file system store and a memory store.
 
 import os
 import tempfile
+from io import BytesIO
 
 from moin.storage.backends.stores import Backend
 from moin.storage.stores.memory import BytesStore as MemoryBytesStore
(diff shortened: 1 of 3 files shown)


roland-ruedenauer:
The contained changes should no be required.

The configuration parameter "create_backend" can be set to **true** and storage folders should be created when the backend is started.

Index creation (cli) should also take care of calling the create method on each configured storage.

Caveat: I didn't test all this - just read the code of _init_backends in app.py

noqt:
Good catch on `create_backend`. A fresh review found that the first version created filesystem paths without knowing whether the backend was read-only.

I've pushed a narrower repair: missing filesystem paths are created only when the owning backend is writable. A new regression opens a read-only backend with both paths absent and proves neither directory is created.

The repaired head passes the focused writable/read-only cases, the full storage suite (108 passed, 12 skipped), Black, Ruff, Bandit, and diff checks. `create_backend` remains the explicit way to create every configured backend; this change only makes opening a newly configured writable filesystem backend safe for its first page.

Built by NOQT.

roland-ruedenauer:
`MoinApp.init_backends` has a boolean parameter `create_backend` controlling whether backends should be just opened or created. The changes in the PR cause ignoring this parameter and will perform storage creation despite the parameter `create_backend` being set to false.

Moin might be changed to always create backends but it is not clear if this is desired. At least this would require further code cleanup (removing the no longer required configuration parameter and argument of `MoinApp.init_backends`).

Then there is still the question if it is sufficient to set the already existing configuration parameter also named `create_backend` to true and achieve auto creation of storage folders.

noqt:
You're right: this patch bypasses the existing `init_backends(create_backend=False)` contract, so I'm withdrawing it rather than defending the broader `open()` behavior. I also traced the suggested alternative: setting the configuration `create_backend = True` does create the missing backend directories through `router.create()`, but the same path then calls `storage.create()`, which recreates the existing Whoosh indexes. That is not a safe answer for an established wiki without a full rebuild. A better follow-up is an explicit backend-only creation path that does not touch indexes. Closing this version keeps the control boundary intact. — NOQT

PR #2375, opened 2026-08-28
Fix 500 on missing archive member in item download

rikvanriel:
Fix 500 on missing archive member in item download

When a user requests /+download/<item>?member=<name> where name is not
present in the tar or zip archive, tarfile.extractfile()/zipfile.open()
raise KeyError, which propagates as an unhandled 500. member is a
user-controlled query parameter, so any name is reachable.

Wrap the member extraction in _do_get and surface a missing member as
400 instead. Adds a regression test that the /+download route returns
400 for a member absent from the archive.

### DIFF
--- a/src/moin/items/_tests/test_Content.py
+++ b/src/moin/items/_tests/test_Content.py
@@ -105,6 +105,37 @@ class TestTarItems:
         item = Item.create(item_name, contenttype="application/x-tar")
         assert item.content.get_member("example1.txt").read() == filecontent
 
+    def test_get_member_missing_raises_keyerror(self):
+        # A member name not present in the archive must raise KeyError
+        # (tarfile.extractfile does) rather than being silently served.
+        item_name = "ContainerItem3"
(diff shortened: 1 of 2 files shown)


roland-ruedenauer:
I would prefer a 400 (Bad Request) in this case.

noqt (approved):
NOQT independently tested current head `5713345e`. This is nicely scoped: the catch sits directly around archive-member lookup, so successful downloads keep their existing path.

Local checks:
- 16 content tests passed, including the new TAR missing-member route case
- the existing frontend download-route test passed
- a separate real ZIP archive probe returned 400 for a missing member
- the same ZIP probe used a hostile-looking member name and confirmed the error page escapes it rather than rendering raw HTML
- `git diff --check` passed

Both TAR and ZIP missing-member lookups raise `KeyError`, so the shared handling covers the two archive implementations without broad exception swallowing. Looks good to me.

UlrichB22:
@rikvanriel, thanks for supporting the Moin2 project.

PR #2376, opened 2026-08-28
Confirm before discarding unsaved edits

### LINKED ISSUE #1564: Show dialog to confirm if user cancels modification
In the current situation if a user clicks on `Cancel` on a modify page the wiki just returns to the previous page. All new changes are gone. It would prevent frustration if the wiki would show a confirmation dialog before it really returns to the previous page.

noqt:
Cancel could discard an editor's unsaved work without warning because each theme cleared the form's changed state before leaving.

This keeps that state until the user chooses what to do. Cancel now asks for confirmation after a real form change, while an unchanged form still closes immediately. The shared handler covers Topside, Focus, and Basic and preserves Basic's existing return-to-item behavior.

The regression suite executes the real shared JavaScript handler and covers rejecting the dialog, accepting it, unchanged forms, Basic navigation, and the Topside proxy button. CI now runs that dependency-free JavaScript test.

Tests:

- 5 JavaScript behavior tests passed
- 78 frontend and theme tests passed
- Black, Ruff, Bandit, and git diff --check passed

Fixes #1564

Built and submitted by NOQT.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,8 @@ jobs:
       run: |
         pip install tox
         tox -e ruff
+    - name: Test JavaScript
+      run: node --test src/moin/static/js/_tests/test_common.js
 
   pytest:
 
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -260,6 +260,11 @@ class TestFrontend:
         cancel_buttons = document.xpath('//*[@id="moin-cancel-text-button"]')
         assert len(cancel_buttons) == 1
         assert "onclick" not in cancel_buttons[0].attrib
+        if theme_name == "topside":
+            proxy_buttons = document.xpath(
+                "//button[@onclick=\"document.getElementById('moin-cancel-text-button').click()\"]"
+            )
+            assert len(proxy_buttons) == 1
 
     def test_modify_item_show_preview(self):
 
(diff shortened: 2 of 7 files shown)


noqt:
I pushed 5bfe606d to close the client-side coverage gap.

The new dependency-free test executes common.js itself and covers rejecting the discard dialog, accepting it, leaving an unchanged form, Basic's direct navigation, and the Topside proxy button. The CI workflow now runs that test. Locally, all 5 JavaScript cases and 78 frontend/theme tests pass, along with Black, Ruff, Bandit, and the diff check.

This follow-up is from NOQT. The normal hosted checks are the next gate before integration.

noqt:
NOQT follow-up on exact head 5bfe606d7bc45e3c6ba7670f55beea0b727aa536:

- The unsaved-edit confirmation now covers reject, accept, unchanged content, Basic navigation, and Topside proxy navigation in 5 JavaScript behavior tests.
- Focused rendered-theme tests pass 3/3; the relevant frontend/theme suite passes 78/78.
- Black, Ruff, Bandit, and diff checks pass.
- The JavaScript behavior test is wired into CI.

We also ran an internal independent review of this exact head and found no blocking correctness, scope, security, or licensing issue. GitHub is not reporting hosted checks yet, so this is local evidence rather than a hosted-CI claim.

I am with NOQT, the contributor. Maintainer review or disposition is welcome.

cintek:
@UlrichB22 I tested this PR and it works.

PR #2377, opened 2026-09-01
Fix crash on invalid regex search terms on Python < 3.13

rikvanriel:
search() caught re.PatternError for an invalid regex term, but that name only exists on Python 3.13+ (a clearer alias for re.error). On Python 3.11 and 3.12, which moin supports, referencing re.PatternError in the except clause itself raised AttributeError -- seen in production as "module 're' has no attribute 'PatternError'".

Use re.error instead, which has always existed and is kept on 3.13+ as an alias.

Adds a regression test mocking the whoosh searcher to raise re.error as the engine does for an invalid regex.


UlrichB22:
@rikvanriel, how can I reproduce the error using the `+search` view with Python 3.12?

roland-ruedenauer:
Try entering e.g. `r"x(a"` into the search text input

PR #2379, opened 2026-09-03
Fix crash on crafted login_persistent param

rikvanriel:
When any page is requested with `?login_persistent=<invalid>`, the auth
setup path calls `parse_bool()` on user-controlled input which raises
`ValueError` for anything outside true/false sets. The exception is
uncaught in `app.py:setup_user → handle_login` → propagates as 500 on
every request, not just login.

Repro: `GET /AnyPage?login_submit=1&login_persistent=maybe` – anonymous,
no auth needed, crashes all pages.

Wrap the `parse_bool` call in `try/except ValueError` and treat invalid
values as `False` (non-persistent login), matching the default when the
param is absent. This prevents user-controlled query input from crashing
the global auth setup path.

Priority: P0 CRITICAL global anon DoS, every page, single query param.
Rebased on top of origin/master e878e94 after #2375 + #2377 merged.

Co-Authored-By: muse-spark-1.2

### DIFF
--- a/src/moin/app.py
+++ b/src/moin/app.py
@@ -374,7 +374,7 @@ def setup_user() -> user.User:
 
     # then handle login/logout forms
     form = request.values.to_dict()
-    if "login_submit" in form:
+    if "login_submit" in form and request.method == "POST":
         # this is a real form, submitted by POST
         userobj = auth.handle_login(userobj, **form)
     elif "logout_submit" in form:
--- a/src/moin/auth/__init__.py
+++ b/src/moin/auth/__init__.py
@@ -423,7 +423,10 @@ def handle_login(userobj: User | None, **kw: Any) -> User | None:
     """
 
     stage = kw.pop("stage", None)
-    persistent = parse_bool(kw.pop("login_persistent", "false"))
+    try:
+        persistent = parse_bool(kw.pop("login_persistent", "false"))
+    except ValueError:
+        abort(400, "Invalid value for login_persistent")
 
     params = {
         "username": kw.pop("login_username", None),


src/moin/auth/__init__.py
```diff
@@ -423,7 +423,12 @@ def handle_login(userobj: User | None, **kw: Any) -> User | None:
     """
 
     stage = kw.pop("stage", None)
-    persistent = parse_bool(kw.pop("login_persistent", "false"))
+    try:
+        persistent = parse_bool(kw.pop("login_persistent", "false"))
+    except ValueError:
+        # login_persistent is user-controlled via query param; an invalid
+        # value must not crash the auth path for every page load
+        persistent = False
```
roland-ruedenauer: IMO this is invalid input and should not be silently ignored; suggest calling abort instead

roland-ruedenauer:
Function `handle_login` should only be called for handling the response of the `login_submit` form. Do we miss a check?

Also, I don't see how this is a critical issue.

rikvanriel:
Sorry, I'm in the middle of some other stuff, so just AI output below. I'm happy to adjust the commit in whatever way you want.

Q1. Should handle_login only be called for the login_submit form? Do we miss a check?​

Yes, that's the deeper issue. app.py:setup_user() does form = request.values.to_dict() where request.values is CombinedMultiDict([args, form]) in Flask – it merges both GET query string and POST form. Then if "login_submit" in form: triggers for any request that has ?login_submit=1 in the query string, not just a real POST from the login template. The comment says "this is a real form, submitted by POST" but the code doesn't enforce request.method == 'POST'.

So GET /AnyPage?login_submit=1&login_persistent=maybe does reach handle_login via setup_user on every page, not just /+login. That's why the parse_bool ValueError crashes every page. The fix in this PR (guard ValueError → False) stops the 500, but the root cause suggests we could also tighten the gate to if request.method == "POST" and "login_submit" in form: to only handle login submissions on POST. I kept this PR minimal to just stop the crash, matching the existing behavior for valid values, but happy to add the POST check as a follow-up or in this PR if preferred.

Q2. Critical?​

Repro: Python 3.11/3.12 (or any version), anonymous, no auth: GET /<any-page>?login_submit=1&login_persistent=maybe → ValueError: Not a boolean value: maybe uncaught in auth/__init__.py:426 → 500 on every request path because setup_user is before_request. Single query param trivially breaks the whole wiki for that request. It's DoS, not data leak, so severity depends on definition – I marked it CRITICAL/P0 because blast radius is "all pages, anonymous, one param" vs other bugs that need a specific endpoint.

And on silent ignore vs abort: suggestion of abort(400) is reasonable – it's invalid input. I chose False to match the existing default when param absent (kw.pop("login_persistent", "false")) and because browsers only send valid checkbox values (on/true), so maybe is clearly a hand-crafted query, not a user error worth a 400 page on every wiki page. A 400 on GET /AnyPage?login_persistent=maybe would also be a DoS-ish UX break (user can't even view page). Silently treating it as "not persistent" is more robust, like how int(request.values.get("w")) failures fall back to None. But willing to switch to abort(400) if project prefers strict validation – let me know and I can update the commit.

roland-ruedenauer:
`handle_login` should clearly only be invoked in cases it was written for (handling login form responses). So the _gate_ should be tightened here. Meaning: checking for method=="POST" and probably the request endpoint too.

Handing of ValueError for `parse_bool` should definitely cause an error response.

The criticality is very low IMO, as the 500 response will only happen for constructed cases ( shit in => shit out ). I'd consider the wording in the problem description as an exaggeration. Nonetheless, a code correction would be very welcome.

src/moin/app.py
```diff
@@ -374,7 +374,11 @@ def setup_user() -> user.User:
 
     # then handle login/logout forms
     form = request.values.to_dict()
-    if "login_submit" in form:
+    if "login_submit" in form and request.method != "POST":
```
roland-ruedenauer: I'd actually prefer not to introduce a special case for a craftet request here and just add the check for request.method == "POST" to the following elif line.

One might also argue, handle_request should only ever by invoked if request.path matches "/+login".

rikvanriel: I changed the code structure as you suggested. That is nicer indeed.

As for your second comment, are you saying that the setup_from_session() already takes care of keeping track of users who are already logged in, and calling handle_request() when users are on another page in the wiki is pointless?

```
    # first try setting up from session
    try:
        userobj = auth.setup_from_session()
    except KeyError:
        # error caused due to invalid cookie, recreating session
        session.clear()
        userobj = auth.setup_from_session()

    # then handle login/logout forms
    form = request.values.to_dict()
    if "login_submit" in form and request.method == "POST":
        # this is a real form, submitted by POST
        userobj = auth.handle_login(userobj, **form)
    elif "logout_submit" in form:
        # currently just a GET link
        userobj = auth.handle_logout(userobj)
    else:
        userobj = auth.handle_request(userobj)
```

What is that handle_request() call supposed to do when the user is not trying to log in, and does not already have a session?

roland-ruedenauer: Thanks for applying the changes. Looks good to me as it is.

My comment simply meant that "login_submit" should currently only be seen when a POST request is sent from path "/+login". I'm not surre why the check here does look for "login_submit" and not the request path.

handle_request() exists as hook for auth-methods to do some per request work.

Again, thanks for your efforts.

PR #2380, opened 2026-09-03
Fix 500 on /+dispatch with missing or unknown endpoint

rikvanriel:
Fix 500 on /+dispatch with missing or unknown endpoint

When a user requests /+dispatch without an endpoint query param, or with
one naming a view that doesn't exist, args.pop("endpoint") or
is_endpoint_expecting() raises KeyError, which propagates as an
unhandled 500. Both are user-controlled inputs, so either is reachable.

Validate that endpoint is present and names a registered view before
forwarding, surfacing missing or unknown endpoints as 400. Adds a
regression test covering the missing, unknown, and valid-endpoint cases.

### DIFF
--- a/src/moin/apps/frontend/_tests/test_frontend.py
+++ b/src/moin/apps/frontend/_tests/test_frontend.py
@@ -283,6 +283,16 @@ class TestFrontend:
         # TODO: Add another test with valid rev1 and rev2 URL args and an existing item.
         self._test_view("frontend.diff", status="404 NOT FOUND", viewargs=dict(item_name="DoesntExist"))
 
+    def test_dispatch_missing_or_unknown_endpoint_returns_400(self):
+        # A missing or unknown endpoint query param must return 400, not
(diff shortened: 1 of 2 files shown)


roland-ruedenauer:
Does anyone know what's the purpose of the "/+dispatch" route? I don't see why this one exists. The endpoint passed in to dispatch could be called directly, if desired!?

Todo: registered routes matching `r"\+.*"` should handle all HTTP methods and never let "+XXX" be treated as a wiki item name by `show_item`. In case of an invalid method an error response should be returned (like e.g. a POST request to "/+dispatch").

UlrichB22:
> Does anyone know what's the purpose of the "/+dispatch" route?

IMO this is one of those functions that are intended “for future use.” I ran a quick test, removed the function, and ran pytest without any problems. Since as usual no one uses these hidden routes, I prefer to remove them. They could become a security risk at some point.

@ThomasWaldmann, @RogerHaase, do you agree?

PR #2382, opened 2026-09-16
Remove unused +dispatch route

UlrichB22:
Related to #2380.


rikvanriel:
Not that I have any say in this, but I'm all for removing a source of 500 errors and stack trace emails that isn't used for anything.

roland-ruedenauer (approved):
LGTM

